Issue #63, opened 2026-08-08
Fix the invalid escape sequence in coordinator.py (SyntaxWarning on Python 3.13)


MSKazemi:
Python 3.13 emits a warning that 3.12 does not:

```
packages/kubeintellect-server/app/agent/nodes/coordinator.py:434
  SyntaxWarning: invalid escape sequence '\`'
    Backslashes (`\`) ARE allowed — they are required for jsonpath separators
```

Now that the runtime image is `python:3.13-slim` (#59), this shows up on every import in the container.

### Why it happens

Line 434 is inside a regular (non-raw) triple-quoted string — part of the coordinator's system prompt. `\`` is not a recognised escape, so Python currently keeps the backslash literally and warns. **In a future Python this becomes a `SyntaxError`**, which is why it is worth fixing now rather than silencing.

### Reproduce

```bash
cd v4
uv run --python 3.13 python -c "import app.agent.nodes.coordinator"
```

### Task

Fix the escape so the intent is explicit and the warning goes away. Either:

- escape it properly: `\\\``, or
- make the string raw (`r"""..."""`) — but **check the whole string first**, because other intentional escapes in that prompt (e.g. `{"\n"}` and `{"\t"}` in the jsonpath examples) would change meaning under a raw string.

The safest fix is usually the targeted one. The prompt text the model sees must not change — that block is telling the model that backslashes are allowed in jsonpath separators, so the rendered output still needs to show a literal backslash.

### Verify

```bash
uv run --python 3.13 python -W error::SyntaxWarning -c "import app.agent.nodes.coordinator"
uv run python -m pytest tests/ -q     # 986 passing, must stay green
```

Then diff the prompt string before and after (print it) and confirm it is byte-identical — this is a prompt, so a silent change to its text is a behaviour change.

Small, self-contained, and a good way to learn why raw strings matter. Comment here to claim it.
See [CONTRIBUTING.md](https://github.com/MSKazemi/kubeintellect/blob/main/CONTRIBUTING.md).

MSKazemi:
Fixed in [`0c7b055`](https://github.com/MSKazemi/kubeintellect/commit/0c7b0555ff40cd47c8b32e33e0b0ded0f5ea0a45).

[...]

PR #73, opened 2026-08-09
fix(ui): resolve eslint errors in PtyTerminal

AdvaitVarhade:
Fixes the 2 ESLint errors in PtyTerminal.tsx as requested in Issue #50.

Please check and tell if any changes are required.

### DIFF
--- a/v4/packages/kube-q/web/app/page.tsx
+++ b/v4/packages/kube-q/web/app/page.tsx
@@ -44,14 +44,16 @@ export default function Home() {
     const saved = typeof window !== "undefined"
       ? window.sessionStorage.getItem(TOKEN_KEY) ?? ""
       : "";
-    setToken(saved || undefined);
-    setTokenReady(true);
+    setTimeout(() => {
+      setToken(saved || undefined);
+      setTokenReady(true);
+    }, 0);
   }, []);
 
   // If the PTY rejected us for auth, prompt for a token.
   useEffect(() => {
     if (status === "error" && /authentication/i.test(statusDetail ?? "")) {
-      setShowTokenPrompt(true);
+      setTimeout(() => setShowTokenPrompt(true), 0);
     }
   }, [status, statusDetail]);
 
--- a/v4/packages/kube-q/web/components/PtyTerminal.tsx
+++ b/v4/packages/kube-q/web/components/PtyTerminal.tsx
@@ -11,7 +11,7 @@
  * Requires the custom Next.js server:  npm run dev:pty
  */
 
-import { forwardRef, useEffect, useImperativeHandle, useRef, useState } from "react";
+import { forwardRef, useEffect, useImperativeHandle, useRef } from "react";
 
 // Dev: set NEXT_PUBLIC_PTY_PORT=3001 (separate pty-server.mjs port).
 // Production: leave unset → connect to /pty-ws on the same origin (server.mjs).
@@ -51,10 +51,8 @@ const PtyTerminal = forwardRef<PtyTerminalHandle, PtyTerminalProps>(
   const containerRef = useRef<HTMLDivElement>(null);
   const termRef = useRef<import("@xterm/xterm").Terminal | null>(null);
   const statusRef = useRef<PtyStatus>("idle");
-  const setStatus = (s: PtyStatus, detail?: string) => {
-    statusRef.current = s;
-    onStatusChange?.(s, detail);
-  };
+  const onStatusChangeRef = useRef(onStatusChange);
+  onStatusChangeRef.current = onStatusChange;
 
   useImperativeHandle(ref, () => ({
     downloadBuffer() {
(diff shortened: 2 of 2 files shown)


MSKazemi:
Merged — thank you, and welcome. Two things you did here were genuinely useful.

First, you found where the errors actually were. Issue #50 reported both
`react-hooks/set-state-in-effect` errors at `PtyTerminal.tsx:47` and `:54`; they were
at those lines in `app/page.tsx`. The issue was wrong and you quietly corrected it.

Second, your `PtyTerminal` change is the right shape and is kept as written. Moving the
notifier inside the connection effect and reaching the callback through a ref means the
effect depends on `authToken` alone, so a parent re-render can no longer tear down a
live PTY session. That fixed the `exhaustive-deps` warning by removing the dependency
rather than by listing it, which is the harder and better answer.

I did rework the `page.tsx` half after merging, in
[183e29e](https://github.com/MSKazemi/kubeintellect/commit/183e29e). Worth explaining,
because the reasoning generalises.

`setTimeout(fn, 0)` doesn't remove a cascading render — it moves it after paint. The
browser now paints one frame with `tokenReady === false`, so the terminal pane flashes
empty on every page load, and the timer is never cleared on unmount. The lint rule goes
quiet while the thing it is pointing at gets slightly worse. Whenever a fix is "same
work, later tick", it's worth asking whether the work belongs there at all.

The issue named both real options, and both applied:

- **Move the update into the callback that causes it.** The auth-failure prompt was an
  effect watching `status`/`statusDetail` — state that `onStatusChange` had *just*
  written. Setting it directly in that callback deletes the effect
  ([`page.tsx#L70-L78`](https://github.com/MSKazemi/kubeintellect/blob/183e29e2487dbb4172c7f63a563c0b3bd6e88dea/v4/packages/kube-q/web/app/page.tsx#L70-L78)).
- **Derive the value during render.** `sessionStorage` is an external store, which is
  exactly what `useSyncExternalStore` is for — and its `getServerSnapshot` argument makes
  it SSR-safe, which a `useState` lazy initialiser can't be. Both the mount effect and
  the `tokenReady` state disappear.

One more, on your ref: `onStatusChangeRef.current = onStatusChange` ran during render.
React may start a render and throw it away, and a ref written there survives the render
that never committed, so the update goes in an effect instead
([caveat](https://react.dev/reference/react/useRef#caveats)). The seed value in
`useRef(onStatusChange)` covers the first mount.

None of this is a knock on the PR — the diagnosis was right and half the change shipped
as you wrote it. You are credited in `CHANGELOG.md` and will be in the release notes.

If you want another: #53 (remaining `mypy` errors) is open and unclaimed, and is the same
kind of task — small, well-bounded, with a gate that tells you when you're done.

MSKazemi:
@AdvaitVarhade — I'd like to invite you to the **Reviewer** rung for `area/kube-q`, specifically the web terminal and demo UI.

[GOVERNANCE.md](https://github.com/MSKazemi/kubeintellect/blob/main/GOVERNANCE.md#reviewers) defines Reviewers as having review authority in **one area** — routed via [CODEOWNERS](https://github.com/MSKazemi/kubeintellect/blob/main/.github/CODEOWNERS), with triage and labelling. No merge rights, no expectation to watch the whole project.

**Why you.** Two things, and the second matters more than the code.

First, your `PtyTerminal.tsx` fix was kept verbatim — moving the notifier inside the connect effect and reaching the current callback through a ref is the correct dependency shape, and it means a re-rendered parent no longer tears down a live PTY session.

Second, **you corrected the issue.** #50 reported both `set-state-in-effect` errors in `PtyTerminal.tsx`; they were in `page.tsx`. Noticing that the report is wrong before writing the fix is the single most useful reviewer habit there is. You also reported the Python 3.13 syntax warnings that became #63, and proposed the `kq export` command.

You are the only contributor who has touched the front end, which makes that area effectively unreviewed today.

**What it would mean in practice**
- Review requests on `v4/packages/kube-q/web/` and the `kq` CLI.
- Saying what you think — including "this is the `setTimeout(fn, 0)` pattern again" when you see it.
- **No obligation, no schedule.** Declining changes nothing about how your contributions are handled.

Interested? Say so here. If the front end isn't where you want to spend time, tell me what is — there are eight fresh docs issues (#86–#93) that need no cluster and no Python at all.

MSKazemi:
Following up on this a while later.

I audited the project's contributor recognition today and checked, person by person, whether
anyone had actually been *told* they were credited. You had not been, and you have been on both
the [Contributors table](https://github.com/MSKazemi/kubeintellect#contributors) and
`.all-contributorsrc` (`code`, `bug`, `ideas`) for some time. So: you are on the list, and I
should have said so at the time.

The specific things it credits you for, since a one-line "thanks" is worth less than the
detail: the `set-state-in-effect` fix here in #73 — where you **corrected the issue itself**,
which had named the wrong file, rather than fixing what you were told to fix — plus proposing
`kq export` and reporting the Python 3.13 syntax warnings.

That last one has had a longer life than you would expect. It became the `SyntaxWarning` CI
gate, `scripts/check-syntax-warnings.py`, which is stdlib-only and independent of ruff on
purpose. Today a contributor added a sibling to it —
[`scripts/check-text-encoding.py`](https://github.com/MSKazemi/kubeintellect/blob/main/scripts/check-text-encoding.py)
— modelled directly on that one, red cases and all. The pattern you set off is now how this
repo holds a whole bug class rather than an instance of it.

Nothing needed from you. I just did not want the audit to end with a file being correct and
nobody being told.

Issue #76, opened 2026-08-10
Sort imports in the 75 files ruff 0.16 flags with I001


MSKazemi:
A self-contained, mostly mechanical slice of the ruff-0.16 backlog. It is a good first issue because the change is generated by a single command, the scope is one rule, and the verification is unambiguous.

## The problem

`ruff` is pinned `<0.16` in `v4/pyproject.toml:25`. Ruff 0.16 enables the import-sorting rule `I001` by default, and `v4/` currently has **75 files with unsorted imports**. Those 75 have to be cleared before the pin can be lifted — and until it is lifted, CI cannot see any rule that 0.16 added.

## Reproduce

```bash
uvx ruff@latest check v4 --select I001 --statistics
# 75  I001  [*] unsorted-imports
```

To see the individual files:

```bash
uvx ruff@latest check v4 --select I001 --output-format concise | cut -d: -f1 | sort -u
```

## What to do

```bash
uvx ruff@latest check v4 --select I001 --fix
```

That is the whole change. It reorders import blocks; it does not rename, add, or remove anything.

**Then confirm nothing broke.** This is the part that matters, and it is why the issue is not purely cosmetic: reordering imports *can* change behaviour when a module depends on import order (circular imports, or an import with a side effect that must run first). It is unlikely here, but "unlikely" is what the test suites are for.

```bash
make setup        # from the repo root — runs all six gates
```

All six must be green: `Lint (ruff)`, `Types (mypy)`, both pytest suites, `File modes`, `Syntax warnings`. `mypy` sits at **0 errors across 171 files**, so if it reports anything at all, it is from this change.

If a file genuinely cannot be reordered, leave it and add `# ruff: isort: skip_file` at the top **with a one-line comment saying why** — then say so in the PR. A justified exception is a fine outcome; a silent one is not.

## Scope — please keep the PR to exactly this

- **Only `I001`.** Do not run a bare `ruff check --fix`; it will pull in other rule families and make the diff unreviewable.
- **⚠️ Do not touch `UP045`.** Ruff's autofix for it rewrites `Optional[X]` → `X | None`, which on this codebase's injected `RunnableConfig` parameters **silently disables RBAC and the human-in-the-loop approval gate while every test still passes.** Details in the tracking issue linked below and in [AGENTS.md](https://github.com/MSKazemi/kubeintellect/blob/main/AGENTS.md) safety invariant #6.
[...]

Issue #56, opened 2026-08-07
bug: the Homebrew formula can't install (nonexistent 1.0.0) and claims MIT for AGPL code


MSKazemi:
> [!IMPORTANT]
> **Correction (2026-08-10) — defect 1 below is FALSE, and defect 3 is mostly false.**
> Verified against live PyPI, and corrected [in this comment](https://github.com/MSKazemi/kubeintellect/issues/56#issuecomment-5245098296).
>
> - **`kube-q` 1.0.0 exists.** It was uploaded `2026-04-10`, making it the *oldest* release. The
>   formula's URL returns HTTP 200 and downloads a real 22,507-byte sdist whose sha256 matches
>   the formula exactly. **`brew install` does not fail at the download step.** The "verification"
>   in the first comment used `sorted(d['releases'])[-3:]`, which prints only the last three
>   entries and structurally cannot show `1.0.0`.
> - **Three of the five hashes are correct** (`kube-q`, `httpx`, `pygments`). The real bug is
>   sharper than "wrong hashes": `rich` and `prompt_toolkit` carry **63-character** values.
>   A sha256 is 64 hex chars, so those were never valid digests and Homebrew rejects them
>   before fetching.
> - **Defect 2 (the licence) is real and unchanged**, and is the main reason this issue stands.
> - **A fourth defect the issue missed:** `homepage` points at `MSKazemi/kube_q`, the wrong
>   repository (the #74/#78 defect). It should be `MSKazemi/kubeintellect`.
>
> The original text is kept below unedited, so the record of what was claimed stays intact.

---

`v4/packages/kube-q/Formula/kube-q.rb` cannot install, and states the wrong licence.

## Three separate defects

**1. It points at a release that does not exist.**

```ruby
url "https://files.pythonhosted.org/packages/source/k/kube-q/kube_q-1.0.0.tar.gz"
```

The earliest version ever published to PyPI is **1.2.0** (`releases: 1.2.0, 1.3.0, 1.4.0, 1.4.1, 1.4.2, 1.4.3`). There is no 1.0.0, so `brew install` fails at the download step. The current source version is **1.5.0** (`v4/packages/kube-q/pyproject.toml`).

**2. The licence is wrong.**

```ruby
license "MIT"
```

[...]

PR #78, opened 2026-08-10
fix(packaging): point PyPI metadata at the canonical repo and adopt PEP 639

### LINKED ISSUE #74: bug: kube-q's PyPI page links to the wrong repository, and ki-protocol's page has no metadata at all
Every link on the [`kube-q` PyPI page](https://pypi.org/project/kube-q/) points at **`MSKazemi/kube_q`** — a different, unmaintained repository — and the [`ki-protocol` PyPI page](https://pypi.org/project/ki-protocol/) has no links, no author, and no classifiers at all.

PyPI is one of this project's main discovery surfaces (`kube-q` serves ~58 downloads/month). Right now a meaningful share of the people who find us there are being sent somewhere we don't read.

## Defect 1 — `kube-q` sends users and bug reports to the wrong repository

`v4/packages/kube-q/pyproject.toml:26-29`:

```toml
[project.urls]
Homepage = "https://github.com/MSKazemi/kube_q"
Repository = "https://github.com/MSKazemi/kube_q"
"Bug Tracker" = "https://github.com/MSKazemi/kube_q/issues"
```

`MSKazemi/kube_q` (underscore) is a **pre-AGPL, pre-`ki-protocol` snapshot**, public, 0 stars, last pushed 2026-07-01. It is not where `kube-q` is developed. The consequences are all live today:

- "Homepage" and "Repository" on PyPI land a prospective user on a stale project.
- **"Bug Tracker" routes bug reports to an issue tracker nobody triages.**
- There is no link to the docs, the website, the changelog, or this repository anywhere on the page.

The canonical repository is `MSKazemi/kubeintellect` — see the note at the top of [CONTRIBUTING.md](https://github.com/MSKazemi/kubeintellect/blob/main/CONTRIBUTING.md).

## Defect 2 — `ki-protocol`'s PyPI page is empty

`v4/packages/ki-protocol/pyproject.toml` declares **no `classifiers`, no `authors`, and no `[project.urls]`**. Both other distributions hard-depend on it (`kube-q` and `kubeintellect` each require `ki-protocol>=1.0.0`), so it is the first thing a resolver pulls — and its page tells a reader nothing, including **no licence classifier**, so PyPI's sidebar shows no licence for an AGPL package.

## Defect 3 — the `license` field uses a form deprecated by PEP 639

All three packages use the table form, and they disagree with each other:

| Package | `license =` | Effect on the published page |
|---|---|---|
| `kube-q` | `{ file = "LICENSE" }` | Dumps the **entire 34 KB AGPL text** into the `License:` metadata field |
| `ki-protocol` | `{ text = "AGPL-3.0-or-later" }` | short string |
| `kubeintellect` | `{ text = "AGPL-3.0-or-later" }` | short string |

None of the three publishes a machine-readable SPDX `License-Expression`. Under [PEP 639](https://peps.python.org/pep-0639/) (supported by the hatchling backend all three already use) the correct form is a SPDX string plus an explicit file list.

## Reproduce

```bash
for p in kube-q ki-protocol kubeintellect; do
  echo "== $p"
  curl -s "https://pypi.org/pypi/$p/json" | python3 -c "
import json,sys; i=json.load(sys.stdin)['info']
print('  urls              :', i.get('project_urls'))
print('  license_expression:', i.get('license_expression'))
print('  license classifier:', [c for c in i['classifiers'] if 'License' in c])"
done
```

## What to do

Metadata-only — no runtime code changes.

1. **`v4/packages/kube-q/pyproject.toml`** — repoint `[project.urls]` at `MSKazemi/kubeintellect`, and match the key set the other packages use so all three PyPI pages look alike. Add `Documentation` and `Changelog`.
2. **`v4/packages/ki-protocol/pyproject.toml`** — add `authors`, `[project.urls]`, and a `classifiers` list including the AGPLv3+ licence classifier and `Programming Language :: Python :: 3.12` / `3.13`, mirroring `kube-q`'s.
3. **All three** — move to the PEP 639 form:

   ```toml
   license = "AGPL-3.0-or-later"
   license-files = ["LICENSE"]
   ```

   and drop the now-redundant `License :: OSI Approved :: ...` classifier (PyPI rejects having both on new uploads).

## Verify

```bash
cd v4
uv build --package kube-q       # and ki-protocol, and kubeintellect
python3 -c "
import zipfile,glob,email
for w in glob.glob('dist/*.whl'):
    m = email.message_from_string(
        zipfile.ZipFile(w).read([n for n in zipfile.ZipFile(w).namelist()
                                 if n.endswith('METADATA')][0]).decode())
    print(w.split('/')[-1])
    print('  License-Expression:', m.get('License-Expression'))
    print('  Project-URL       :', m.get_all('Project-URL'))
"
```

Every wheel should report `License-Expression: AGPL-3.0-or-later` and a `Project-URL` set pointing at `MSKazemi/kubeintellect`. Then run the normal gates (`make setup` from the repo root runs all six).

## Notes for whoever takes this

- **Fixing the source does not fix the live pages.** PyPI metadata is immutable per release, so these pages stay wrong until the next publish. That is expected — land the fix, and it ships with the next release.
- **Do not delete or alter the `MSKazemi/kube_q` repository as part of this.** It is a live public repo with its own publishing workflow, and it is out of scope here. This issue only changes where *our* metadata points.
- Please comment before starting so nobody duplicates the work.

shaurya703:
Fixes #74

## What & why

Metadata-only, per the issue — no runtime code changes.

1. **`kube-q`** — `[project.urls]` no longer points at the stale `MSKazemi/kube_q` snapshot. All keys now target `MSKazemi/kubeintellect`, with `Documentation`, `Changelog` (kube-q's own changelog file in this repo), and `Issues` added so the key set matches the other two packages.
2. **`ki-protocol`** — adds `authors`, `[project.urls]`, and a `classifiers` list (mirroring kube-q's, including `Programming Language :: Python :: 3.12` / `3.13`), so its PyPI page is no longer empty.
3. **All three** — moved to the PEP 639 form (`license = "AGPL-3.0-or-later"` + `license-files = ["LICENSE"]`) and dropped the now-redundant `License :: OSI Approved :: ...` classifier, since PyPI rejects new uploads carrying both. The licence is machine-readable via `License-Expression` instead.

Two supporting additions: `LICENSE` (byte-identical to the repo-root copy) is added to `ki-protocol/` and `kubeintellect-server/`, because `license-files = ["LICENSE"]` resolves relative to each package directory — previously neither wheel shipped the licence text at all. The kubeintellect-server URLs also gain `Repository`/`Changelog` and the canonical `MSKazemi` casing, so all three pages look alike.

## Verify

Ran the issue's verification snippet after `uv build --package` for each; every wheel reports:

```
License-Expression: AGPL-3.0-or-later
License-File      : ['LICENSE']
Project-URL       : Homepage/Documentation/Repository/Changelog/Issues -> MSKazemi/kubeintellect
License classifiers: []
```

## Type of change

- [x] 🐛 Bug fix

## Scope

- Version directory touched: v4/

## Checklist

- [ ] New behavior has both a **happy-path** and an **error-path** test (n/a — packaging metadata only)
- [x] `uv run pytest` passes locally (server: 995 passed; kq: 312 passed)
- [x] `uv run mypy src` passes locally (0 errors)
- [ ] Docs updated if behavior/CLI/flags changed (n/a)
- [x] Commits are signed off (`git commit -s` — DCO)

## Notes for reviewers

- All six `make setup` gates are green locally.
- As the issue notes, the live PyPI pages stay wrong until the next publish; this only fixes the source metadata.
- `MSKazemi/kube_q` itself is untouched, per the issue's out-of-scope note.
- Observed but deliberately left alone (outside this issue's scope): `v4/packages/kube-q/mkdocs.yml` still carries `repo_url`/`site_url` pointing at the old `kube_q` repo — happy to file or fix separately if wanted.

### DIFF
--- a/v4/packages/ki-protocol/LICENSE
+++ b/v4/packages/ki-protocol/LICENSE
@@ -1,0 +1,661 @@
+                    GNU AFFERO GENERAL PUBLIC LICENSE
+                       Version 3, 19 November 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
+  The GNU Affero General Public License is a free, copyleft license for
+software and other kinds of works, specifically designed to ensure
+cooperation with the community in the case of network server software.
+
+  The licenses for most software and other practical works are designed
+to take away your freedom to share and change the works.  By contrast,
+our General Public Licenses are intended to guarantee your freedom to
+share and change all versions of a program--to make sure it remains free
+software for all its users.
+
+  When we speak of free software, we are referring to freedom, not
+price.  Our General Public Licenses are designed to make sure that you
+have the freedom to distribute copies of free software (and charge for
+them if you wish), that you receive source code or can get it if you
+want it, that you can change the software or use pieces of it in new
+free programs, and that you know you can do these things.
+
+  Developers that use our General Public Licenses protect your rights
+with two steps: (1) assert copyright on the software, and (2) offer
+you this License which gives you legal permission to copy, distribute
+and/or modify the software.
+
+  A secondary benefit of defending all users' freedom is that
+improvements made in alternate versions of the program, if they
+receive widespread use, become available for other developers to
+incorporate.  Many developers of free software are heartened and
+encouraged by the resulting cooperation.  However, in the case of
+software used on network servers, this result may fail to come about.
+The GNU General Public License permits making a modified version and
+letting the public access it on a server without ever releasing its
+source code to the public.
+
+  The GNU Affero General Public License is designed specifically to
+ensure that, in such cases, the modified source code becomes available
+to the community.  It requires the operator of a network server to
+provide the source code of the modified version running there to the
+users of that server.  Therefore, public use of a modified version, on
+a publicly accessible server, gives the public access to the source
+code of the modified version.
+
+  An older license, called the Affero General Public License and
+published by Affero, was designed to accomplish similar goals.  This is
+a different license, not a version of the Affero GPL, but Affero has
+released a new version of the Affero GPL which permits relicensing under
+this license.
+
+  The precise terms and conditions for copying, distribution and
+modification follow.
+
+                       TERMS AND CONDITIONS
+
+  0. Definitions.
+
+  "This License" refers to version 3 of the GNU Affero General Public License.
+
+  "Copyright" also means copyright-like laws that apply to other kinds of
+works, such as semiconductor masks.
+
+  "The Program" refers to any copyrightable work licensed under this
+License.  Each licensee is addressed as "you".  "Licensees" and
+"recipients" may be individuals or organizations.
+
+  To "modify" a work means to copy from or adapt all or part of the work
+in a fashion requiring copyright permission, other than the making of an
+exact copy.  The resulting work is called a "modified version" of the
+earlier work or a work "based on" the earlier work.
+
+  A "covered work" means either the unmodified Program or a work based
+on the Program.
+
+  To "propagate" a work means to do anything with it that, without
+permission, would make you directly or secondarily liable for
+infringement under applicable copyright law, except executing it on a
+computer or modifying a private copy.  Propagation includes copying,
+distribution (with or without modification), making available to the
+public, and in some countries other activities as well.
+
+  To "convey" a work means any kind of propagation that enables other
+parties to make or receive copies.  Mere interaction with a user through
+a computer network, with no transfer of a copy, is not conveying.
+
+  An interactive user interface displays "Appropriate Legal Notices"
+to the extent that it includes a convenient and prominently visible
+feature that (1) displays an appropriate copyright notice, and (2)
+tells the user that there is no warranty for the work (except to the
+extent that warranties are provided), that licensees may convey the
+work under this License, and how to view a copy of this License.  If
+the interface presents a list of user commands or options, such as a
+menu, a prominent item in the list meets this criterion.
+
+  1. Source Code.
+
+  The "source code" for a work means the preferred form of the work
+for making modifications to it.  "Object code" means any non-source
+form of a work.
+
+  A "Standard Interface" means an interface that either is an official
+standard defined by a recognized standards body, or, in the case of
+interfaces specified for a particular programming language, one that
+is widely used among developers working in that language.
+
+  The "System Libraries" of an executable work include anything, other
+than the work as a whole, that (a) is included in the normal form of
+packaging a Major Component, but which is not part of that Major
+Component, and (b) serves only to enable use of the work with that
+Major Component, or to implement a Standard Interface for which an
+implementation is available to the public in source code form.  A
+"Major Component", in this context, means a major essential component
+(kernel, window system, and so on) of the specific operating system
+(if any) on which the executable work runs, or a compiler used to
+produce the work, or an object code interpreter used to run it.
+
+  The "Corresponding Source" for a work in object code form means all
+the source code needed to generate, install, and (for an executable
+work) run the object code and to modify the work, including scripts to
+control those activities.  However, it does not include the work's
+System Libraries, or general-purpose tools or generally available free
+programs which are used unmodified in performing those activities but
+which are not part of the work.  For example, Corresponding Source
+includes interface definition files associated with source files for
+the work, and the source code for shared libraries and dynamically
+linked subprograms that the work is specifically designed to require,
+such as by intimate data communication or control flow between those
+subprograms and other parts of the work.
+
+  The Corresponding Source need not include anything that users
+can regenerate automatically from other parts of the Corresponding
+Source.
+
+  The Corresponding Source for a work in source code form is that
+same work.
+
+  2. Basic Permissions.
+
+  All rights granted under this License are granted for the term of
+copyright on the Program, and are irrevocable provided the stated
+conditions are met.  This License explicitly affirms your unlimited
+permission to run the unmodified Program.  The output from running a
+covered work is covered by this License only if the output, given its
+content, constitutes a covered work.  This License acknowledges your
+rights of fair use or other equivalent, as provided by copyright law.
+
+  You may make, run and propagate covered works that you do not
+convey, without conditions so long as your license otherwise remains
+in force.  You may convey covered works to others for the sole purpose
+of having them make modifications exclusively for you, or provide you
+with facilities for running those works, provided that you comply with
+the terms of this License in conveying all material for which you do
+not control copyright.  Those thus making or running the covered works
+for you must do so exclusively on your behalf, under your direction
+and control, on terms that prohibit them from making any copies of
+your copyrighted material outside their relationship with you.
+
+  Conveying under any other circumstances is permitted solely under
+the conditions stated below.  Sublicensing is not allowed; section 10
+makes it unnecessary.
+
+  3. Protecting Users' Legal Rights From Anti-Circumvention Law.
+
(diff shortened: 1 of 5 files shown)


MSKazemi:
Thanks @shaurya703 — picking this up now, workflow run approved.

Reviewing properly and will come back shortly. Flagging up front that I'll look closely at two things: the `license-files` relative-path reasoning behind adding the two `LICENSE` copies, and whether dropping the `License ::` classifier is safe for the *existing* published versions as well as new uploads.

Good catch on `mkdocs.yml` — that's the same defect as the issue, and I'd rather it were fixed than filed.

MSKazemi (approved):
Reviewed and verified. This is better than what the issue asked for, in a way worth naming.

## You fixed something the issue missed

The issue said to add `license-files = ["LICENSE"]`. It did not notice that **the path resolves relative to each package directory**, and that `ki-protocol` and `kubeintellect-server` had no `LICENSE` file of their own. So the instruction as written would have produced a build error or a silently missing licence.

You caught that and added the two copies. I verified they are byte-identical to the repo root:

```
v4/packages/ki-protocol/LICENSE            IDENTICAL to root LICENSE
v4/packages/kubeintellect-server/LICENSE   IDENTICAL to root LICENSE
```

The consequence is bigger than the issue: **neither wheel was shipping the licence text at all.** For AGPL-licensed code distributed as a binary artifact, that is a real compliance gap, not a cosmetic one — and it was invisible from the source tree. Both wheels and all three sdists now carry it:

```
ki_protocol-1.0.0.dist-info/licenses/LICENSE
kube_q-1.5.0.dist-info/licenses/LICENSE
kubeintellect-2.2.0.dist-info/licenses/LICENSE
```

## Verified against real built artifacts

I built all three rather than reading the TOML, because the question "what will PyPI actually show" is only answerable from the metadata:

| | `Metadata-Version` | `License-Expression` | `License-File` | legacy `License:` | `License ::` classifier |
|---|---|---|---|---|---|
| `ki-protocol` 1.0.0 | 2.4 | `AGPL-3.0-or-later` | `LICENSE` | absent | none |
| `kube-q` 1.5.0 | 2.4 | `AGPL-3.0-or-later` | `LICENSE` | absent | none |
| `kubeintellect` 2.2.0 | 2.4 | `AGPL-3.0-or-later` | `LICENSE` | absent | none |

The legacy `License:` field being **absent** on `kube-q` is the specific defect from the issue: `license = { file = "LICENSE" }` was dumping the entire 34 KB AGPL text into that header. It is now a single SPDX expression.

Every `Project-URL` on all three now resolves to `MSKazemi/kubeintellect`, and both `Changelog` targets exist (`CHANGELOG.md`, `v4/packages/kube-q/CHANGELOG.md`).

**Will PyPI accept it?**

```
$ twine check dist/*
Checking dist/ki_protocol-1.0.0-py3-none-any.whl: PASSED
Checking dist/kube_q-1.5.0-py3-none-any.whl: PASSED
Checking dist/kubeintellect-2.2.0-py3-none-any.whl: PASSED
  … and all three sdists: PASSED
```

Dropping the `License ::` classifier is correct and required — PyPI rejects an upload carrying both a `License-Expression` and a licence classifier.

**And it still installs and runs**, which is the check that actually matters:

```
$ pip install dist/ki_protocol-*.whl dist/kube_q-*.whl && kq --version
kube-q 1.5.0
```

All nine `kq` subcommands still resolve — probed **without** `--help`, since argparse's `-h` exits 0 even when a subcommand is missing.

All nine required checks are green on `f34075f8`, and `make check-modes` / `make check-syntax` pass on the merge result.

## Your `mkdocs.yml` note — I'd rather fix it than file it, and it goes further than you saw

You flagged `v4/packages/kube-q/mkdocs.yml` still pointing at the old repo. You were right, and when I swept for it there are **13 files** in `v4/` with the same defect. The one that matters most:

**`v4/packages/kube-q/README.md` is `readme = "README.md"` in the pyproject — it *is* the body of the PyPI page.** It currently tells readers `git clone https://github.com/MSKazemi/kube_q`. So this PR fixes the links in the sidebar while the page body underneath still sends people to the wrong repository.

That is not a gap in your PR — you fixed precisely what #74 described, and the issue only described the metadata. I am taking the doc sweep on as separate maintainer work rather than growing your PR, so yours stays exactly what it claims to be. Your find is credited in the changelog entry for it.

Nothing to change here. Thank you — reading `license-files` carefully enough to notice the relative-path problem is the kind of care that turns a metadata patch into a compliance fix.

MSKazemi:
Merged, and the doc sweep from your `mkdocs.yml` note is now on `main` too (`06a774b`).

It came to 13 files. The one worth knowing about: **`v4/packages/kube-q/README.md` is `readme` in the package's pyproject, so it *is* the body of the PyPI page** — and it was telling readers to `git clone https://github.com/MSKazemi/kube_q`. Your PR fixed the sidebar links; without the sweep the page underneath would still have pointed people elsewhere. So that one-line observation was worth more than it looked.

Two things the sweep turned up that I would not have predicted:

- **`edit_uri` had to move with `repo_url`.** It is relative, so once `repo_url` became the monorepo, `edit/main/docs/` pointed at a path that does not exist — every "edit this page" link would have 404'd. It now carries the package path, verified 200.
- **The mkdocs social link to `ghcr.io/mskazemi/kube_q` was already dead** (404 — that package does not exist). Repointed at the image that does.

I left `site_url` alone deliberately: `mskazemi.github.io/kube_q` redirects to a **live** docs site built from the old repo's `gh-pages`, so moving it is an owner decision about where docs get published, not a link fix.

The new from-source instructions were verified end to end rather than assumed — clone the monorepo, `cd kubeintellect/v4/packages/kube-q`, `pip install -e .`, and `kq --version` reports 1.5.0. Worth noting that path only resolves at all now that `ki-protocol` is on PyPI; it would have failed a week ago.

Both of your PRs are in, #74 and #76 are closed, and CI is green on `main`. Two contributions in one afternoon, both landed unmodified — thank you. If you want another, #56 (the Homebrew formula) is the natural next one, and it needs a judgement call rather than just a patch: the right answer may well be to delete it.

MSKazemi:
@shaurya703 — I'd like to invite you to the **Reviewer** rung for `area/packaging`.

[GOVERNANCE.md](https://github.com/MSKazemi/kubeintellect/blob/main/GOVERNANCE.md#reviewers) defines Reviewers as contributors trusted with review authority in **one area**. It is deliberately not co-maintaining the whole project: review requests for your area get routed to you via [CODEOWNERS](https://github.com/MSKazemi/kubeintellect/blob/main/.github/CODEOWNERS), you can triage and label issues, and you do **not** need or take on merge rights.

**Why packaging, and why you.** #74 asked you to repoint some URLs and adopt PEP 639. You did that — and then noticed that `license-files` resolves relative to each package directory, which meant `ki-protocol` and `kubeintellect-server` had **no `LICENSE` of their own and had been shipping wheels with no licence text at all.** In an AGPL project that is a compliance defect, it was invisible from the source tree, and my issue had missed it entirely. You also flagged `mkdocs.yml`, which turned out to be 13 files with the same defect — including the one that *is* the PyPI page body.

That is exactly the instinct the rung is for: reading the thing the instruction actually depends on, rather than the instruction.

**What it would mean in practice**
- You get review requests on `v4/packages/*/pyproject.toml`, the Homebrew formula, `snap/`, and the publish workflows.
- Reviewing means saying what you think — "this looks right", "this claim isn't verified", "I'd check X". You are never the last line of defence; I still merge.
- **No obligation and no schedule.** Say no, or say "not right now", and nothing changes about how your PRs are treated. Stepping back later is normal and documented.

If you're interested, say so here and I'll add you to CODEOWNERS for that path. If packaging isn't what interests you, tell me which area is — #56 (Homebrew: fix or delete, your call) is open and is genuinely a judgement call rather than a patch.

Issue #93, opened 2026-08-10
docs: add a worked example for `kq completion`


MSKazemi:
`v4/docs/examples.md` is the cookbook — nine worked scenarios, each showing a real command and its real output. **It documents only 2 of the 10 `kq` subcommands.** `kq completion` is one of the eight with no worked example.

Verified on `main` (2026-08-10):

```console
$ grep -c 'kq completion' v4/docs/examples.md
0
```

## Why this matters

A two-minute quality-of-life win that nobody discovers.

The command is in `kq help` and in `v4/docs/cli-reference.md`, so people know it exists — they just have no idea what it looks like in use. A reference entry tells you the flags; a worked example tells you whether you want it.

## What to do

Add a numbered section to `v4/docs/examples.md` following the existing house style (see `## 6 · Survey cluster health` for the shape):

- a one-line statement of the situation the reader is in
- the exact command
- **real output** in a ```console block
- two or three sentences on how to read that output
- a link to the relevant deeper doc where one exists

`completion` — *Print a shell completion script (bash, zsh, or fish)*.

## Acceptance criteria

- [ ] `grep -c 'kq completion' v4/docs/examples.md` returns ≥ 1
- [ ] The output shown is **real**, produced by actually running the command — not written by hand
- [ ] It matches the numbering and heading style of the existing sections
- [ ] `v4/docs/examples.md` § Related still links correctly
- [ ] No claim in the new section contradicts `v4/docs/cli-reference.md`

## How to test

```bash
pip install kube-q            # 1.5.0 or newer
kq completion --help               # flags
kq completion                      # then run it for real and copy the actual output
```

⚠️ **Paste real output, and redact node names, namespaces and registries before committing.**
If the command needs a session id or a cluster you do not have, say so on this issue rather than
inventing plausible output — a fabricated example is worse than a missing one, and it is the one
failure mode this project explicitly cannot ship.

## Task metadata

| | |
|---|---|
| **Difficulty** | Beginner |
| **Estimated time** | 30–60 minutes |
| **Skills** | Markdown. Running one CLI command. **No Python, no cluster required** if you use the hosted read-only API (`kq --api-key ki-ro-dev`) |
| **Mentor** | @MSKazemi — ask on this issue, or in [Q&A](https://github.com/MSKazemi/kubeintellect/discussions/categories/q-a) |
| **Status** | 🟢 **Unclaimed** — comment "I'd like to take this" and it will be assigned to you |

[...]

PR #105, opened 2026-08-10
docs: document data handling boundaries

### LINKED ISSUE #82: Document what data leaves the cluster, what is stored, and what is redacted
There is no document in this repository that states **what cluster data leaves the cluster, what is persisted, and what is redacted.** That is the first thing a platform team asks before evaluating an AI tool with cluster access, and today the only way to find out is to read `app/utils/redact.py` and the tool layer.

This surfaced while answering [a Q&A discussion](https://github.com/MSKazemi/kubeintellect/discussions/83) about data egress.

## Why it matters more than a typical docs gap

The facts are genuinely reassuring in places and genuinely limited in others, and **the current silence lets a reader assume either one.** Specifically, it is easy to see that `app/utils/redact.py` exists and conclude the project filters what it sends to the model. It does not — that redactor is applied on the way **into storage**, not on the way **out to the provider**. Its own docstring says so:

> Secret/PII redaction for stored reflexion outcomes… Anything that lands in the database must pass through this redactor.

Call sites are `memory/episodes.py`, `memory/preferences.py`, `db/flight_recorder.py`, and `tools/kubectl_tool.py` (pre-state capture). None of them is the prompt path.

## What the document should state plainly

1. **What is sent to the model provider** — `kubectl` output (names, images, events, conditions), Prometheus series labels and values, and Loki log lines. Name Loki explicitly as the highest-risk source, because application logs contain whatever the application prints.
2. **What can never be read, and therefore never sent** — Secrets and ServiceAccounts are blocked at the tool layer *for every role including `superadmin`* (`KUBECTL_BLOCKED_RESOURCES`). This is a structural guarantee, not a filter, and it deserves to be stated as such.
3. **What is persisted, where, and for how long** — episodes, the temporal knowledge graph, the flight recorder, preferences. What redaction applies to each, and that it is heuristic (the module says "full enumeration of secret formats is impossible").
4. **What is *not* redacted** — the live prompt path. Say it explicitly rather than leaving it inferable.
5. **How to reduce exposure** — `readonly` keys, not wiring Loki, scoping the ServiceAccount, `rbac.allowExec: false`.
6. **Self-hosted inference** — `OPENAI_BASE_URL` accepts OpenAI-compatible endpoints, but this is not a tested or supported path today (#17). Do not imply it is.
7. **Telemetry** — whether Langfuse tracing, when enabled, sends prompt content anywhere, and what the default is.

## Where it should live

A new `v4/docs/data-handling.md`, linked from `README.md`, `SECURITY.md`, `v4/docs/security.md` and `SUPPORT.md`. `security.md` today covers authentication, roles, RBAC tiers, the HITL gate and secret protection — it stops short of data flow, which is a different question.

## Scope note

**This is a documentation issue, not a request to build a redaction layer.** An egress filter for prompts is a legitimate and much larger discussion (and would need its own RFC); the job here is to describe accurately what exists today. If writing it makes a gap obvious, file that separately rather than growing this issue.

Accuracy matters more than completeness here — please verify each claim against the code rather than against this issue body, and say so in the PR. Getting this document subtly wrong is worse than not having it.

---

## Task metadata

| | |
|---|---|
| **Difficulty** | Intermediate — accuracy matters more than volume |
| **Skills** | Markdown, plus enough Python to read `app/utils/redact.py` and the tool layer and verify each claim against the code. |

**Verify every statement against the code, not against the issue body.** Getting this document subtly wrong is worse than not having it.

uuzzrm:
## Summary

- add a v4 data-handling page covering model inputs, persistence, redaction boundaries, and telemetry
- link it from the root README, SECURITY.md, SUPPORT.md, v4 security docs, and the MkDocs navigation
- document the current Secret/ServiceAccount blocklist, heuristic storage redaction, checkpoint/KG limitations, Langfuse default, and self-hosted provider status

Closes #82

## Verification

- `make check-modes`
- `make check-syntax`
- `cd v4 && uv run mkdocs build`
- `cd v4 && uv run python scripts/check_doc_claims.py`
- `cd v4 && uv run --frozen ruff check packages/kubeintellect-server/app/ packages/ki-protocol/`
- `cd v4 && uv run --frozen mypy packages/kubeintellect-server/app packages/ki-protocol packages/kube-q/kube_q`
- `cd v4 && uv run --frozen python -m pytest tests/ -q` — 995 passed
- `cd v4/packages/kube-q && uv run --frozen python -m pytest tests/ -q` — 310 passed, 2 unrelated failures in CLI/Rich rendering and pager behavior

Strict MkDocs mode still reports the repository's existing missing PDF target
from `v4/docs/research.md`; the non-strict build succeeds.

## AI assistance

I used AI assistance to inspect the repository and. I
reviewed the final text against the current v4 source paths listed in the new
page and ran the checks above. This PR changes documentation only; it does not
add an egress-redaction layer.

Signed-off-by: Ruiming Zhao <uuzzrm@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,7 @@ Run `make help` at the root at any time to see the available infra targets.
 
 - **[Website](https://kubeintellect.com/)** — overview, live demo, and hosted API.
 - **[v4 docs](v4/docs/)** — install, quickstart, configuration, architecture, security, CLI reference, and troubleshooting for the current version.
+- **[Data handling](v4/docs/data-handling.md)** — what v4 sends to model and telemetry endpoints, stores, and redacts.
 - **[v4 README](v4/README.md)** — every install path in detail.
 - Each version directory (`v1/`–`v4/`) ships its own `README.md` and `docs/`.
 
@@ -168,8 +169,9 @@ Known and deliberate, so you can judge fit before installing:
 - **An LLM API key is required.** OpenAI or Azure OpenAI out of the box; v4 also
   supports Anthropic, Qwen, and OpenAI-compatible endpoints. Queries cost tokens.
 - **Cluster context leaves your network** unless you point it at a self-hosted or
-  in-cluster model endpoint. Review [SECURITY.md](SECURITY.md) before running it
-  against production.
+  in-cluster model endpoint. Review [SECURITY.md](SECURITY.md) and the
+  [data-handling notes](v4/docs/data-handling.md) before running it against
+  production.
 - **It is not a replacement** for your observability stack, a GitOps/CD pipeline,
   or on-call judgment. It queries the tools you already run.
 - **LLM answers can be wrong.** The approval gate exists precisely because the
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -38,6 +38,9 @@ Security-relevant areas we especially care about:
 - **Privilege escalation** — a tool obtaining Kubernetes permissions beyond least-privilege.
 - **Sandbox escape** — for any generated/dynamic code execution path.
 
+For the separate question of what cluster data can enter model prompts,
+telemetry, and persistent stores, see [Data handling in v4](v4/docs/data-handling.md).
+
 ## Out of scope
 
 - Vulnerabilities in third-party dependencies should generally be reported upstream (tell us too if
--- a/SUPPORT.md
+++ b/SUPPORT.md
@@ -24,6 +24,7 @@ Most first-run problems are covered already:
 
 - **[v4 README](v4/README.md)** — every install path (browser, CLI-only, local Kind, Docker Compose, existing cluster)
 - **[v4 docs](v4/docs/)** — install, quickstart, configuration, architecture, security, CLI reference
+- **[Data handling](v4/docs/data-handling.md)** — model inputs, persistence, redaction, and telemetry boundaries
 - **[Troubleshooting](v4/docs/troubleshooting.md)** and **[FAQ](v4/docs/faq.md)**
 - **[ROADMAP.md](ROADMAP.md)** — including what the project deliberately *will not* do
 
--- a/v4/docs/data-handling.md
+++ b/v4/docs/data-handling.md
@@ -1,0 +1,181 @@
+# Data handling in KubeIntellect v4
+
+Read this before connecting KubeIntellect to a production cluster. The server
+is the trust boundary: it can read cluster state, put selected results into an
+LLM request, and persist several kinds of operational history. This page
+describes the current v4 implementation. It is not a promise that arbitrary
+application output is safe to send to a provider.
+
+## The short version
+
+- The model can receive the user's request, the v4 system prompt, cluster
+  snapshots, and results returned by the tools.
+- `kubectl` output is passed through with size limits, not a general-purpose
+  secret scrubber. Prometheus results include series labels and values. Loki
+  results include stream labels and log lines; application logs are the
+  highest-risk input because an application can print arbitrary content.
+- With the default resource blocklist, `Secret` and `ServiceAccount` resources
+  are rejected by the kubectl tool before kubectl is called. This check also
+  applies to `superadmin`; that role only bypasses the infrastructure-namespace
+  write block.
+- The redactor is primarily a storage safeguard. It is not an egress filter for
+  the live prompt, tool results, or telemetry callbacks.
+- Langfuse tracing is off by default. When enabled and configured, treat the
+  configured Langfuse host as an additional recipient of trace data.
+
+## What can reach an LLM provider
+
+The coordinator builds a request from the system prompt, the current
+conversation, memory context, and the current cluster snapshot. Tool calls can
+add more context before a response is produced. Depending on which tools are
+used, that context can contain:
+
+| Source | Examples of data returned to the agent | Important boundary |
+|---|---|---|
+| `run_kubectl` | Resource names, images, namespaces, conditions, events, descriptions, and `kubectl logs` output | Output is capped, but is not generally redacted before it becomes model context. Namespace listings have a separate blocked-namespace filter. |
+| `query_prometheus` | PromQL text, metric labels, current values, and min/average/max values for range queries | No secret redaction is applied to labels or values. |
+| `query_loki` | LogQL text, stream labels, timestamps, and log lines | Treat every returned log line as potentially sensitive or instruction-like. Loki is optional and returns an error when `LOKI_URL` is unset. |
+| Conversation and memory | User text, remembered preferences, session notes, failure hints, recalled episodes, and recent KG changes | These are assembled into the live prompt when the corresponding memory features are enabled. |
+
+The tool and prompt paths do apply size limits (`kubectl` 8,000 characters;
+Prometheus and Loki 6,000 characters; coordinator tool messages 2,000
+characters). A size limit controls volume, not sensitivity, and does not make
+the remaining text trustworthy.
+
+The server sends the resulting LLM request to the configured provider. If
+`PROMETHEUS_URL` or `LOKI_URL` points outside the cluster, the server also sends
+queries to those configured endpoints; the returned data then follows the
+model-context path above.
+
+## What the kubectl tool blocks
+
+The default configuration is:
+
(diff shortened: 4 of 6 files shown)


MSKazemi (approved):
Reviewed. **Every factual claim in this page checks out against the source**, including the parts I expected to find wrong. Thank you — this is the hardest kind of documentation to get right, and #82 said explicitly that getting it subtly wrong would be worse than not having it.

## What I verified, claim by claim

I did not spot-check. For a page that tells people what protects their cluster data, every number and behavioural statement was read against the code:

| Claim in the page | Source | ✓ |
|---|---|---|
| kubectl output cap 8,000 chars | `kubectl_tool.py:567` `limit = 8_000` | ✅ |
| Prometheus / Loki caps 6,000 | `prometheus_tool.py:22`, `loki_tool.py:24` `_OUTPUT_CAP = 6_000` | ✅ |
| Coordinator tool message cap 2,000 | `coordinator.py:117` `_TOOL_OUTPUT_MAX_CHARS = 2_000` | ✅ |
| `LANGFUSE_ENABLED` defaults false | `config.py:123` | ✅ |
| `REFLEXION_REDACT_SECRETS` defaults true, 1,500-char cap | `config.py:166`, `flight_recorder.py:44` | ✅ |
| Token frames skipped by the recorder | `flight_recorder.py:41` `_SKIP_KINDS = frozenset({"token"})` | ✅ |
| Preference values redacted, capped 300 | `preferences.py:64` | ✅ |
| Inferred-preference decay default 60 days | `config.py:274` `PREFERENCE_DECAY_DAYS = 60` | ✅ |
| RCA outcomes 90 days, low-confidence patterns 30 | `schema.sql:134-135` | ✅ |
| KG write path has **no** redactor call | `memory/kg.py` — 0 occurrences | ✅ |
| Checkpoint path has **no** redactor call | `agent/workflow.py` — 0 occurrences | ✅ |
| Loki returns an error when `LOKI_URL` is unset | `loki_tool.py:53-54` | ✅ |
| `rbac.allowExec: false` default; local dev enables it | `values.yaml:38` false, `values-local.yaml:7` true | ✅ |

**The claim I most expected to be overstated was the one about the blocklist, and it is exactly right.** `_check_protected_access(verb, args)` takes **no role parameter**, so role-independence is structural rather than a policy someone could regress. It is called at `kubectl_tool.py:468` — before the HITL interrupt at `:478` and before `subprocess.run` at `:528`. "Before kubectl is called and before an approval prompt" is precisely true.

You also got the `superadmin` nuance right, which is the subtle bit: the bypass re-runs the check and only clears the error when the target resource is not itself blocklisted.

## The one correction — and it is not your error

I pushed one commit (`cee5cd6`, with you as `Co-authored-by`) narrowing a single sentence.

The page said `superadmin` bypasses the *"namespace **write** restriction"*. It bypasses the whole namespace check. `_check_protected_access`'s own docstring says **"All operations on infrastructure namespaces are blocked, including reads"**, and the bypass clears that error for **any verb** whenever the resource is not blocklisted — so a `superadmin` can *read* `kube-system` too, not only write to it.

**You reproduced the wording of the code's own comment**, which says "namespace write block" at `kubectl_tool.py:466` and "no namespace write restrictions" at `:447`. The code comment is the thing that is imprecise; I am correcting that separately on `main` so the next reader is not misled the same way.

In a page whose whole purpose is telling an operator what is and is not protected, understating a privileged role's reach is worth the extra sentence.

## Your test report was right, and it found a real bug in our suite

You wrote: *"310 passed, 2 unrelated failures in CLI/Rich rendering and pager behavior"*. That was accurate and it was worth reporting rather than quietly ignoring. On the merge result I get **312 passed** — so nothing in your change caused it. Chasing why, on `main`:

```console
$ COLUMNS=60 pytest tests/ -q   →  1 failed, 311 passed   (test_findings_table)
$ COLUMNS=40 pytest tests/ -q   →  5 failed, 307 passed
$ TERM=dumb  pytest tests/ -q   →  1 failed, 311 passed
$ NO_COLOR=1 pytest tests/ -q   →  1 failed, 311 passed
$ (default)  pytest tests/ -q   →  312 passed
```

**The `kq` suite inherits terminal width, `TERM` and `NO_COLOR` from the environment.** CI happens to run at a width where it passes, which is why nobody noticed. That is a defect that makes contributors think they broke something — filed as **#106**, with your report credited as the source.

## Gates, run on the final tree

My push moved the head, so these are on `cee5cd6` merged into `main`, not on what I first reviewed:

| Gate | Result |
|---|---|
| `make check-modes` / `check-syntax` | OK (299 files) |
| `ruff check` (pinned, as CI runs it) | All checks passed |
| `mypy` | 0 errors, 171 files |
| server suite | **995 passed** |
| `kq` suite | **312 passed** |
| `mkdocs build` | exit 0 |
| `scripts/check_doc_claims.py` | *"Doc claims match the code"* |

All nine required checks are green on `cee5cd6`.

On the strict-mode note in your description — you were right that the missing PDF target in `v4/docs/research.md` is pre-existing and not yours. Thanks for separating that out instead of "fixing" it in this PR.

## On the AI-assistance disclosure

Exactly what [CONTRIBUTING § AI assistance](https://github.com/MSKazemi/kubeintellect/blob/main/CONTRIBUTING.md#ai-assistance) asks for, and you did the thing that actually matters: you listed the source paths the claims rest on, which is what made this reviewable in an hour instead of a day. The `## Source paths checked` section at the bottom of the page should stay — it tells the next person how to re-verify when the code moves.

MSKazemi:
Merged as `a60dbd0` — [`v4/docs/data-handling.md`](https://github.com/MSKazemi/kubeintellect/blob/main/v4/docs/data-handling.md) is live, and #82 closed with it.

**Where you're credited**, so you can see it rather than take my word for it:

- [README contributors table](https://github.com/MSKazemi/kubeintellect#contributors) — named, with what the page actually does
- `.all-contributorsrc` (`doc` + `security`)
- [#106](https://github.com/MSKazemi/kubeintellect/issues/106) — the environment-sensitivity bug your test report uncovered, filed with your report as the source
- The next release notes

You are also now on the repository's Contributors graph. There were four humans on it this morning; you are the fifth.

**Two things, and neither is an expectation.** Your contribution is complete and nothing below is owed:

1. If the page was useful to write — or if the project is useful to you — a ⭐ genuinely helps. At this size it is one of the few signals other engineers use to decide whether a young project is worth an evening, and we are at the stage where single digits move it.
2. If you ever want to write about it, say so and I will point people at *you*. I would rather amplify a contributor than ask a contributor to amplify the project — you did the work, and the interesting story here is a data-handling page whose every claim was verified against source, not a repo asking for attention.

And if neither appeals, that is completely fine. The thing I actually want is the thing you already did.

If you want another: **#106** is yours by right if you want it — you found it. Otherwise **#56** (Homebrew: fix it or delete it, genuinely your call) or any of [#86–#93](https://github.com/MSKazemi/kubeintellect/labels/good%20first%20issue), which need no cluster and no Python.

MSKazemi:
Correction to my message above, and an apology — I got a fact wrong.

I said **#106** was "yours by right if you want it". It isn't: @floze-the-genius had already claimed it at 17:10 with a concrete plan, *before* I wrote that. I should have checked the issue before offering it, and I'm sorry — being offered something and then having it withdrawn is worse than never being offered it.

**What stands unchanged:** #106 exists because *your* test report was honest. You could easily have written "310 passed" and moved on; instead you flagged two failures as unrelated, and that turned out to be a real environment-sensitivity bug that makes contributors think they broke something. You are credited as the source in the issue, and that credit is not affected by who fixes it.

If you'd like another, all of these are live and unclaimed:

- **#56** — the Homebrew formula. The most interesting one on the list, because the right answer may well be *delete it*, and that judgement is genuinely yours to make.
- **[#86–#93](https://github.com/MSKazemi/kubeintellect/labels/good%20first%20issue)** — eight worked CLI examples, one per subcommand. No cluster, no Python.
- **#82's neighbours** — if the data-handling page raised anything you thought was missing while writing it, that observation is worth an issue of its own. You have read that part of the codebase more carefully than anyone else has.

And the standing offer from my previous comment holds: if you write about any of this, tell me and I'll point people at you.

Issue #22, opened 2026-08-02
Add a bounded retry to triage JSON parsing


MSKazemi:
When the model returns malformed JSON in the triage step (`v4/packages/kubeintellect-server/app/cortex/graph.py`), there's currently no retry — it falls straight through. A bounded retry/repair before fallback would improve robustness.

### Scope
- Add a small bounded retry/repair in the triage node.
- A test that feeds malformed-then-valid JSON.

uuzzrm:
Picked this up — opened #133. It adds a bounded repair loop in the triage node: if the reply doesn't parse as JSON, the malformed output goes back with a corrective hint and we retry (up to 3 calls total) before falling back to investigate. Also split the parser so the loop can distinguish a bad reply from a deliberate `investigate` answer. Happy to adjust the attempt count or the hint wording if you'd rather have it as a setting.

MSKazemi:
@uuzzrm — you asked this on 14 August and I never answered it. Eight days, on a direct question, from the person who then went and built the thing. That is on me, and it is worth answering properly now that #133 is merged.

> Happy to adjust the attempt count or the hint wording if you'd rather have it as a setting.

**Keep it a constant.** `_TRIAGE_MAX_PARSE_ATTEMPTS = 3` is right where you put it, and here is the distinction I want the codebase to hold:

`CORTEX_MAX_GATHER_ROUNDS` **is** a setting (default 8, `app/core/config.py:286`) because it governs a loop whose cost is legitimately a *cluster operator's* trade-off — more rounds means more evidence and more tokens, and the right number genuinely differs between a laptop kind cluster and a busy production namespace.

The parse-repair bound is not that. It is a **safety bound on a provider defect**: the model returned something that is not the JSON it was asked for. There is no operator preference hiding in that number. Exposing it would create exactly two new failure modes and no new capability — set to `1` it silently disables the repair you built (and the issue quietly regresses to the pre-#133 behaviour with no signal that it did), set to `50` it spends fifty calls arguing with a model that is not going to comply. A knob whose only settings are "break it" and "waste money" is a liability, not flexibility.

The hint wording I would leave alone too. It names the schema, forbids prose and fences, and echoes the bad reply back — capped at 2,000 chars, which is the detail I want to point at, because it is the part most people would have left unbounded. A pathological reply re-sent on every remaining attempt is a context-window exhaustion bug that only shows up in production, and you closed it before it existed.

Two things I noticed reading the merged code, neither a change request:

[...]

Issue #14, opened 2026-08-02
Make the memory recall similarity noise-floor configurable


MSKazemi:
The trigram recall similarity noise-floor is currently a hard-coded module global:

- `v4/packages/kubeintellect-server/app/memory/episodes.py` — `_TRGM_FLOOR = 0.02`
- `v4/packages/kubeintellect-server/app/memory/summaries.py` — `_SUMMARY_FLOOR = 0.02`

### Task
Promote it to a setting in `app/core/config.py` (keep `0.02` as the default) and read it from settings in both modules. Add/adjust a test.

Self-contained, teaches the config pattern, no cluster needed.

---

## Task metadata

| | |
|---|---|
| **Difficulty** | Beginner |
| **Estimated time** | 1–2 hours |
| **Skills** | Python. Reading an existing pydantic-settings class and following the pattern. No cluster, no API key — the suites are mocked. |
| **Mentor** | @MSKazemi — ask on this issue, or in [Q&A](https://github.com/MSKazemi/kubeintellect/discussions/categories/q-a) |
| **Status** | 🟢 **Unclaimed** — comment "I'd like to take this" and it will be assigned to you |

uuzzrm:
I would like to take this. I will keep the default floor unchanged, expose it through the v4 settings, and add focused tests for the configured threshold on both recall paths.

MSKazemi:
Closing a loop I should have closed at the time — you claimed this on 12 August and I never replied here.

Merged as [#116](https://github.com/MSKazemi/kubeintellect/pull/116): `MEMORY_RECALL_SIMILARITY_FLOOR` in `app/core/config.py`, defaulting to the existing `0.02`, read by both `episodes.py` and `summaries.py` — exactly the scope you proposed. Thanked properly on the PR itself, just not here, where you actually asked.

PR #107, opened 2026-08-10
test(kube-q): pin terminal env for deterministic Rich output

### LINKED ISSUE #106: bug: the kq test suite fails depending on terminal width, TERM and NO_COLOR
The `kube-q` test suite inherits the terminal size from the environment, so it passes or fails depending on how wide the contributor's terminal happens to be. A contributor sees failures that look like their fault, in tests unrelated to their change.

**Found via [#105](https://github.com/MSKazemi/kubeintellect/pull/105).** @uuzzrm reported "310 passed, 2 unrelated failures in CLI/Rich rendering and pager behavior". That was accurate and honest — and it is a bug in our suite, not in their change.

## Measured on `main` (2026-08-10)

```console
$ uv run python -m pytest tests/ -q                     # default terminal
312 passed

$ COLUMNS=60 uv run python -m pytest tests/ -q
1 failed, 311 passed        # tests/test_findings_cmd.py::test_findings_table

$ COLUMNS=40 uv run python -m pytest tests/ -q
5 failed, 307 passed

$ TERM=dumb uv run python -m pytest tests/ -q
1 failed, 311 passed

$ NO_COLOR=1 uv run python -m pytest tests/ -q
1 failed, 311 passed
```

So the suite is sensitive to **width, `TERM`, and `NO_COLOR`** — three things that differ on every contributor's machine and are not pinned anywhere. CI happens to run at a width where it passes, which is exactly why this went unnoticed.

## Why it matters more than a flaky test normally would

A first-time contributor runs the suite, sees red in files they never touched, and has no way to know it is not their fault. The most likely outcomes are that they spend an hour chasing it, or quietly abandon the PR. `CONTRIBUTING.md` tells them a green local run is the bar, so the failure reads as a blocker.

## What to do

Pin the rendering environment for the whole suite in `v4/packages/kube-q/tests/conftest.py`, so the tests assert against a deterministic console rather than the developer's terminal:

- force a fixed width (Rich reads `COLUMNS`, and `Console(width=...)` where a console is constructed directly)
- pin `TERM` and colour behaviour so `NO_COLOR` / `TERM=dumb` do not change output
- apply it as an autouse fixture so no individual test has to remember

`tests/test_render.py` is the only file that references width/console/pager directly today, but the failures at `COLUMNS=40` show more than that file is affected.

## Acceptance criteria

- [ ] `COLUMNS=40`, `COLUMNS=60`, `COLUMNS=200`, `TERM=dumb`, and `NO_COLOR=1` all give **312 passed**
- [ ] No test assertion is weakened to achieve that — pin the environment, do not loosen the expectation
- [ ] A one-line comment in `conftest.py` explaining why the pin exists, so nobody removes it later

## How to test

```bash
cd v4/packages/kube-q
for e in "COLUMNS=40" "COLUMNS=60" "COLUMNS=200" "TERM=dumb" "NO_COLOR=1"; do
  echo -n "$e -> "; env $e uv run python -m pytest tests/ -q 2>&1 | tail -1
done
```

## Task metadata

| | |
|---|---|
| **Skills** | Python, pytest fixtures. Some familiarity with Rich helps but the fix is mostly a `conftest.py` autouse fixture. No cluster, no API key. |

AshSgDe29071999:
## Summary
- Add an autouse pytest fixture in `v4/packages/kube-q/tests/conftest.py` that pins `COLUMNS=120`, `TERM=xterm-256color`, and clears `NO_COLOR`.
- Keeps assertions intact while making the suite independent of the local terminal.

Closes #106

Made with [Cursor](https://cursor.com)

### DIFF
--- a/v4/packages/kube-q/tests/conftest.py
+++ b/v4/packages/kube-q/tests/conftest.py
@@ -1,0 +1,17 @@
+"""Shared pytest fixtures for kube-q.
+
+Pin terminal rendering env for the whole suite. Rich/console output otherwise
+depends on the developer's COLUMNS, TERM, and NO_COLOR, which makes unrelated
+tests fail on narrow or dumb terminals (see #106).
+"""
+
+from __future__ import annotations
+
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def _pin_terminal_rendering_env(monkeypatch: pytest.MonkeyPatch) -> None:
+    monkeypatch.setenv("COLUMNS", "120")
+    monkeypatch.setenv("TERM", "xterm-256color")
+    monkeypatch.delenv("NO_COLOR", raising=False)


MSKazemi:
Thank you for this, and I'm sorry — you hit a collision that was my fault, not yours.

#106 had already been claimed on the issue about an hour before this PR opened, and I had confirmed it there. I should have labelled it `claimed` the moment I replied so it was visible from the issue list; I did not, so you had no way of knowing. That is a process failure on my side, and it cost you real work.

**Your diagnosis was correct, and your fix is not wasted.** I want to be concrete rather than consoling:

I ran your branch against the full width/TERM/NO_COLOR matrix on a clean checkout:

```
COLUMNS=40   -> 312 passed
COLUMNS=60   -> 312 passed
COLUMNS=200  -> 312 passed
TERM=dumb    -> 312 passed
NO_COLOR=1   -> 1 failed  (tests/test_config.py::test_cmd_show_runs)
```

Four of the five hostile environments fixed, with no assertion weakened — you found the right root cause and reached for the right mechanism. The remaining case is a genuinely subtle one: an autouse fixture runs *after* pytest imports the test modules, and by then the module-level Rich console has already been constructed with the `NO_COLOR`-stripped theme, so the fixture is too late to affect it. It needs a `pytest_configure` hook to run before collection.

**Your PR is what established that.** I did not know the fixture-only approach was insufficient until I had your branch to run as the control against #109's. The autouse fixture you wrote also survives in the merged version — #109 keeps one, for the same reason you added it, so individual tests can still override with `monkeypatch`. You will be credited in the `CHANGELOG.md` entry for exactly that.

Closing this only because #106 is now fixed on `main` (#109), not because of anything wrong with your work.

**If you'd like another, these are unclaimed right now and none of them will collide** — comment on whichever appeals and I'll label it yours immediately:

- [#22](https://github.com/MSKazemi/kubeintellect/issues/22) — add a bounded retry to triage JSON parsing. Python, closest to what you just did.
- [#95](https://github.com/MSKazemi/kubeintellect/issues/95) — a liveness-probe-failure playbook. No Python, no cluster required.
- [#14](https://github.com/MSKazemi/kubeintellect/issues/14) — make the memory recall similarity noise-floor configurable.

I have also started labelling claimed issues `claimed` so this cannot happen to the next person. That is a direct result of your PR, and I'd rather say so than let it look like nothing came of it.

If the project looks useful, a star helps other engineers find it, and passing it to someone who runs Kubernetes helps more — both optional, and neither has any bearing on your contributions here. Thanks again for the careful work.

MSKazemi:
Following up on this, because a closed PR is the easiest place for a project to quietly forget
someone.

I audited the contributor roster today. You were in the README Contributors table but
**missing from `.all-contributorsrc`** — the machine-readable surface, so the credit existed in
prose and nowhere a tool could see it. Fixed on `main`
([#169](https://github.com/MSKazemi/kubeintellect/pull/169),
[`762a26f`](https://github.com/MSKazemi/kubeintellect/commit/762a26f5511074713efeeaa461d2426952d1b3e2)),
listed as `ideas` and `test`.

**Listed, even though this PR did not merge.** That is deliberate and I want to say why plainly
rather than let it look like a consolation. A roster records the contribution that arrived, not
whether it happened to land. Yours arrived, it was correct, and it collided with an issue
claimed an hour earlier that I had failed to label — a process failure on my side that cost you
real work.

And it is not only a fairness argument. Running your branch as a control is the **only** reason
we know the `pytest_configure` hook in #109 is load-bearing rather than incidental: your
fixture-only version fixed four of the five hostile environments and left `NO_COLOR` failing,
which is what pinned down that the module-level Rich console is constructed at import time,
before any fixture can run. Without your branch to compare against, #109 would have looked like
one arbitrary implementation among several. It is the difference between a fix that works and a
fix that is understood.

The `claimed` label exists because of the collision you hit. It is in use now.

Nothing needed from you — and if you want another one, the `good first issue` queue is real and
I label claims immediately these days.

Issue #96, opened 2026-08-10
playbook: in-cluster DNS resolution failing


MSKazemi:
The detector engine ships **18 declarative failure playbooks**. This proposes one more: **In-cluster DNS resolution failing**.

A playbook is a single YAML file — `detect → investigate → remediate`. **No Python.** It is the lowest-barrier way to make the product measurably better, and it is the lane that scales, because Kubernetes has far more failure modes than 18.

## Why this one

DNS is the single most common cause of "the service is up but my app can't reach it", and it currently has no playbook. It also produces misleading application-level errors, so people debug the wrong layer for a long time.

## Where it goes

`v4/packages/kubeintellect-server/app/agent/playbooks/dns_resolution_failure.yaml`

Read `service_no_endpoints.yaml` first — it is the clearest example of the full schema:

```yaml
name: <PascalCaseName>
detect:
  watch_predicates:      # kind + reason_regex over the kubectl --watch sensorium
  promql:                # optional; a zero-token signal
  debounce_seconds:      # avoid firing on transient states
triggers:                # event_reason_regex / event_message_regex
investigation_steps:     # ordered, concrete kubectl commands with what to look for
expected_evidence:       # what confirms the diagnosis
recommended_fix_template: |
  # what to tell the operator — NOT executed automatically
```

## What to put in it

**Signals to detect:** Application logs containing `no such host`, `Temporary failure in name resolution`, or `dial tcp: lookup ... on 10.96.0.10:53`; CoreDNS pods not `Running` or logging `SERVFAIL`

**Notes:** Cover both directions: CoreDNS itself unhealthy, and a workload with a broken `dnsPolicy`/`dnsConfig` or a `NetworkPolicy` blocking egress to port 53.

## Acceptance criteria

- [ ] The file parses: `python3 -c "import yaml,sys; yaml.safe_load(open('v4/packages/kubeintellect-server/app/agent/playbooks/dns_resolution_failure.yaml'))"`
- [ ] `name` is unique across the directory and PascalCase
- [ ] `investigation_steps` are **real commands**, in the order a human would actually run them
- [ ] `expected_evidence` states what would *confirm* the diagnosis — and, where relevant, what absence of evidence means
- [ ] `recommended_fix_template` **describes** the fix; it must not assume anything is auto-applied
- [ ] `debounce_seconds` is set so a transient state does not fire it
- [ ] Both suites still pass

## How to test

```bash
make setup                    # six of the nine required checks; no cluster, no API key
[...]

PR #108, opened 2026-08-10
feat(playbooks): add NetworkPolicy silently blocking traffic playbook

### LINKED ISSUE #99: playbook: NetworkPolicy silently blocking traffic
The detector engine ships **18 declarative failure playbooks**. This proposes one more: **NetworkPolicy silently blocking traffic**.

## Why this one

NetworkPolicies fail closed and produce **no events at all**. Connections simply time out, which is the hardest failure mode in this list to attribute — and there is no playbook for it.

## Where it goes

`v4/packages/kubeintellect-server/app/agent/playbooks/networkpolicy_blocking.yaml`

```yaml
name: <PascalCaseName>
detect:
recommended_fix_template: |
```

## What to put in it

**Signals to detect:** Connection timeouts between pods where the Service has healthy endpoints; a `NetworkPolicy` selecting the target pods; traffic working after the policy is removed in a test namespace

**Notes:** The signal is the *absence* of evidence — Service and endpoints are healthy, DNS resolves, and the connection still hangs. Say that explicitly in `expected_evidence`.

## Acceptance criteria

- [ ] The file parses: `python3 -c "import yaml,sys; yaml.safe_load(open('v4/packages/kubeintellect-server/app/agent/playbooks/networkpolicy_blocking.yaml'))"`
- [ ] Both suites still pass

## How to test

```bash
```

## Task metadata

| | |
|---|---|

Priyanshu608:
Closes #99

Added playbook for NetworkPolicy silently blocking traffic in v4/packages/kubeintellect-server/app/agent/playbooks/networkpolicy_blocking.yaml.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,25 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
   path now builds its message as a text node instead of assigning `innerHTML`.
 
 ### Added
+- **A `NetworkPolicyBlocking` playbook** (#99, #108) — the 19th, contributed by
+  [@Priyanshu608](https://github.com/Priyanshu608). It covers the one failure in the set that
+  emits **no evidence at all**: a NetworkPolicy denial is discarded in the CNI datapath, so the
+  API server records no event and no Warning is ever produced. The Service is healthy, endpoints
+  exist, DNS resolves — and the connection simply hangs. The playbook makes that *absence* the
+  signal, which is what none of the other 18 could express.
+
+  Two maintainer corrections landed on top. The submitted trigger also matched
+  **`connection refused`**, which is evidence *against* a policy drop: a refusal means a TCP RST
+  came back, so the packet was delivered. Matching it would have pointed the agent at a
+  NetworkPolicy for what is almost always a wrong port or a missing endpoint —
+  `ServiceUnreachable`'s territory. It now matches timeout signatures only, and
+  `test_networkpolicy_blocking_ignores_connection_refused` locks that in. The `triggers:` block
+  was also nested under `detect:`, where the loader never reads it
+  ([`loader.py`](v4/packages/kubeintellect-server/app/agent/playbooks/loader.py) takes
+  `triggers` from the top level), so the playbook loaded cleanly into the registry and then
+  never fired — inert, with no error anywhere. `test_each_playbook_has_complete_schema` catches
+  exactly this and did.
+
 - **The test suites now run on Python 3.13 as well as 3.12** (`Tests (… · py3.13)`), closing
   a gap where the project was verified on an interpreter it does not ship. `v4/Dockerfile`'s
   runtime stage is `python:3.13-slim` and all three distributions declare
--- a/v4/docs/agent-behaviors.md
+++ b/v4/docs/agent-behaviors.md
@@ -123,7 +123,7 @@ detects a matching pattern in the snapshot, the coordinator's system prompt
 includes the playbook(s) inline — guiding it to follow proven steps before
 improvising.
 
-**Playbooks shipped (18):**
+**Playbooks shipped (19):**
 
 *Pod / container lifecycle*
 
@@ -150,6 +150,7 @@ improvising.
 - `JobBackoffLimitExceeded`
 - `ServiceNoEndpoints` (selector / label drift)
 - `ServiceUnreachable`
+- `NetworkPolicyBlocking` (policy drops traffic silently — no event is ever emitted)
 - `WebhookAdmissionRejected`
 
 **Schema** (drop a YAML file into `app/agent/playbooks/`):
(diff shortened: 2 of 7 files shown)


MSKazemi (approved):
Thank you — merging this. You picked the hardest failure mode in the set and got the insight right; I fixed the wiring and pushed it to your branch rather than sending it back, so the credit and the commit stay yours.

**What you got right, and it is the part that mattered.** This line is the reason the playbook is worth having:

> "Absence of evidence: NetworkPolicies fail closed silently, so no Kubernetes warning events will be recorded for dropped connections."

None of the other 18 playbooks can express that. Every one of them keys off something *present* — a status, a reason, an event. You wrote the one whose signal is that nothing is there, and you documented it as a positive finding rather than a gap. That framing survives into the merged version almost verbatim.

## What I changed, and why

**1. The playbook could never fire.** `triggers:` was nested inside `detect:`, but the loader reads `triggers` from the top level ([`loader.py:72`](https://github.com/MSKazemi/kubeintellect/blob/b14e2d92201f909b538c170bf979549aee1bd55b/v4/packages/kubeintellect-server/app/agent/playbooks/loader.py#L72-L75)), and `parse_detect_block` returns `None` when a `detect:` block has no predicates. So it loaded into the registry cleanly, logged no error, and then matched nothing — `match_playbooks` skips any playbook with empty triggers:

```python
>>> pb = _load_one(Path("networkpolicy_blocking.yaml"))
>>> pb.triggers
()          # match_playbooks() skips it
>>> pb.detect
None        # the detector engine never sees it
```

Inert, with no failure anywhere. This is the exact class of bug the project cares most about, so I'm glad it was caught by a gate and not by a user: `test_each_playbook_has_complete_schema` fails on it with `NetworkPolicyBlocking has no triggers`. That test is why this could not have reached `main` quietly — worth knowing it exists.

**2. I dropped `connection refused` from the trigger.** This one is a genuine Kubernetes networking subtlety rather than a style call. A NetworkPolicy denial *drops* the packet — nothing comes back, so the caller hangs and eventually times out. `connection refused` means a TCP RST was returned, which means the packet **was** delivered and something actively rejected it. It is evidence *against* a policy block, and it is almost always a wrong port or a missing endpoint — `ServiceUnreachable`'s territory, which already triggers on that exact string. Matching it here would have sent the agent looking for a NetworkPolicy on the one signal that rules one out.

There is now a regression test pinning this, and I checked it actually bites — restoring your original regex makes it fail:

```
AssertionError: assert 'NetworkPolicyBlocking' not in
    ['NetworkPolicyBlocking', 'ReadinessProbeFailing', 'ServiceUnreachable']
```

**3. `detect: null`, with the reason written down.** Your `watch_predicates: []` was the honest instinct — there is nothing to watch — and `detect: null` is how the schema says that explicitly (same as `command_hardcoded_failure.yaml`). I added a comment recording *why*: the drop happens in the CNI datapath, so no event ever reaches the API server, and drop counters are CNI-specific (`cilium_drop_count_total`, `calico_denied_packets`) rather than guaranteed cluster state.

**4. Two additions to the investigation steps** from the same reasoning: check the **source** namespace for its own egress policy (easy to miss, and blocks identically), and confirm endpoints exist so `ServiceNoEndpoints` is ruled out first.

**5. Docs.** Adding a playbook moves a number that appears in four documentation files, and the repo has a gate for exactly that — `test_doc_claims.py` failed with six mismatches until they were all updated. Playbook count 18 → 19 across `agent-behaviors.md`, `capabilities.md`, `glossary.md` and `architecture.md`, plus a CHANGELOG entry crediting you. Nothing you were expected to know about; noting it so the mechanism isn't a surprise next time.

## Verification

All on the merge tree (`a8a3d47`), not on your original commit:

```
server suite   998 passed          (995 + the 3 tests added here)
kube-q suite   312 passed
ruff           All checks passed
mypy           Success: no issues found in 171 source files
doc claims     pass
```

All ten required checks green, including both Python 3.13 jobs and the install smoke test.

Your commit stays in the history as the feature, with mine on top as the fix, and you're credited in `CHANGELOG.md` and the release notes.

If you'd like another, [#94](https://github.com/MSKazemi/kubeintellect/issues/94) (PVC stuck Pending) and [#96](https://github.com/MSKazemi/kubeintellect/issues/96) (in-cluster DNS failing) are unclaimed and are the same shape — you now know the schema and the trigger-placement trap, which is most of the work. Comment on either and I'll label it yours.

If KubeIntellect looks useful, a star helps other engineers find it, and passing it to someone who runs Kubernetes helps more than a star does — both entirely optional. Thanks for a genuinely good first contribution.

MSKazemi:
Following up on this one.

I audited the project's contributor recognition today. You were credited in the README
Contributors table but were **missing from `.all-contributorsrc`** — the machine-readable
surface, which is what any tool reading this repo's contributor data actually sees. So the
credit existed in prose and nowhere a tool could find it.

Fixed on `main` ([#169](https://github.com/MSKazemi/kubeintellect/pull/169),
[`762a26f`](https://github.com/MSKazemi/kubeintellect/commit/762a26f5511074713efeeaa461d2426952d1b3e2)):
you are listed as `code`, alongside your existing
[Contributors table](https://github.com/MSKazemi/kubeintellect#contributors) row.

While I was in there — the thing about `NetworkPolicyBlocking` that has stayed with me is that
it is the only playbook in the set whose signal is the **absence** of evidence. A policy denial
is dropped in the CNI datapath, so no event is ever emitted and the connection simply hangs.
Every other playbook keys off something that appears. Yours had to key off something that never
will, and you captured that as a positive finding rather than writing it up as a gap. That is a
harder thing to get right than it looks, and it is why the playbook is useful.

Nothing needed from you — I just did not want the roster fix to land silently.

PR #111, opened 2026-08-10
fix(packaging): update kube-q Homebrew formula

### LINKED ISSUE #56: bug: the Homebrew formula can't install (nonexistent 1.0.0) and claims MIT for AGPL code

uuzzrm:
## What & why

Closes #56.

The formula still pointed at the removed `kube-q` 1.0.0 source archive, declared MIT even though the current package is AGPL, and shipped only part of the runtime dependency set. This updates the formula to the published `kube-q` 1.5.0 release, uses the canonical repository homepage and license, and declares the complete runtime resource tree required by Homebrew's dependency-free pip installation. `certifi` is provided by Homebrew and is explicitly excluded from the PyPI resources.

## Type of change

- [x] 🐛 Bug fix

## Scope

## Validation

- `git diff --check`
- Downloaded the main sdist and all 18 declared resources, verified every SHA-256 against the downloaded bytes, and checked the PyPI metadata for the declared URLs and Homebrew's 24-hour release cooldown.
- Resolved the active recursive runtime dependency graph for Python 3.12: 19 packages are covered by the formula resources or the Homebrew `certifi` dependency.
- `python -m pip install --dry-run --ignore-installed --report - kube-q==1.5.0` — 20 packages resolved.
- Reviewed the formula against the current Homebrew `virtualenv_install_with_resources`, `std_pip_args`, `update-python-resources`, and official Python formula patterns.

Homebrew itself, Ruby, and WSL are not installed in the development environment, so `brew audit --strict` and `brew install --build-from-source` could not be run locally. I am reporting that limitation explicitly rather than treating the static and PyPI checks as an install test.

## Notes for reviewers

`pydantic-core` is built from its sdist and therefore includes the `maturin` and Rust build dependencies. PyYAML's source build uses the Homebrew `libyaml` dependency.

## AI assistance

Substantial AI assistance was used to inspect the published package metadata. I reviewed the resulting diff and ran the validation listed above; the Homebrew-native checks remain for CI or a maintainer environment because Homebrew is unavailable locally.

### DIFF
--- a/v4/packages/kube-q/Formula/kube-q.rb
+++ b/v4/packages/kube-q/Formula/kube-q.rb
@@ -2,31 +2,107 @@ class KubeQ < Formula
   include Language::Python::Virtualenv
 
   desc "Interactive terminal CLI for chatting with your Kubernetes cluster via an AI backend"
-  homepage "https://github.com/MSKazemi/kube_q"
-  url "https://files.pythonhosted.org/packages/source/k/kube-q/kube_q-1.0.0.tar.gz"
-  sha256 "0ecce3a15800343c3cbcd9490ce695e923b4390c0808629fcaec5ea03a9c967d"
-  license "MIT"
+  homepage "https://github.com/MSKazemi/kubeintellect"
+  url "https://files.pythonhosted.org/packages/92/85/5d0af5d4fcff49447325ffd338c2ef5e386d99b6a106b570f8e97bb66d60/kube_q-1.5.0.tar.gz"
+  sha256 "738363d6f3c5c1f97c561924ab232e2cfc0ca6cacb5c9bdb18ab756aa1b1cfc1"
+  license "AGPL-3.0-or-later"
 
+  depends_on "certifi" => :no_linkage
+  depends_on "libyaml"
+  depends_on "maturin" => :build
   depends_on "python@3.12"
+  depends_on "rust" => :build
+
+  pypi_packages exclude_packages: "certifi"
+
+  resource "anyio" do
+    url "https://files.pythonhosted.org/packages/61/cc/a381afa6efea9f496eff839d4a6a1aed3bfafc7b3ab4b0d1b243a12573dd/anyio-4.14.2.tar.gz"
+    sha256 "cfa139f3ed1a23ee8f88a145ddb5ac7605b8bbfd8592baacd7ce3d8bb4313c7f"
+  end
+
+  resource "annotated-types" do
+    url "https://files.pythonhosted.org/packages/5f/56/a8120250d128bed162cd73c76d45f6ef9991f3e068f62a8ee060afa3104a/annotated_types-0.8.0.tar.gz"
+    sha256 "13b2beaad985e05e2d6407ee4c4f35590b11f8d693a258a561055cac8f64cab7"
+  end
+
+  resource "h11" do
+    url "https://files.pythonhosted.org/packages/01/ee/02a2c011bdab74c6fb3c75474d40b3052059d95df7e73351460c8588d963/h11-0.16.0.tar.gz"
+    sha256 "4e35b956cf45792e4caa5885e69fba00bdbc6ffafbfa020300e549b208ee5ff1"
+  end
+
+  resource "httpcore" do
+    url "https://files.pythonhosted.org/packages/06/94/82699a10bca87a5556c9c59b5963f2d039dbd239f25bc2a63907a05a14cb/httpcore-1.0.9.tar.gz"
+    sha256 "6e34463af53fd2ab5d807f399a9b45ea31c3dfa2276f15a2c3f00afff6e176e8"
+  end
 
   resource "httpx" do
-    url "https://files.pythonhosted.org/packages/source/h/httpx/httpx-0.28.1.tar.gz"
+    url "https://files.pythonhosted.org/packages/b1/df/48c586a5fe32a0f01324ee087459e112ebb7224f646c0b5023f5e79e9956/httpx-0.28.1.tar.gz"
     sha256 "75e98c5f16b0f35b567856f597f06ff2270a374470a5c2392242528e3e3e42fc"
   end
 
-  resource "rich" do
-    url "https://files.pythonhosted.org/packages/source/r/rich/rich-14.3.3.tar.gz"
-    sha256 "5a5c2a4d5e8bbf946abbc1c86e8a69a7e0fb70d8e8c41e6aeeb3d08b7d6c4c3"
+  resource "idna" do
+    url "https://files.pythonhosted.org/packages/cd/63/9496c57188a2ee585e0f1db071d75089a11e98aa86eb99d9d7618fc1edce/idna-3.18.tar.gz"
+    sha256 "ffb385a7e039654cef1ab9ef32c6fafe283c0c0467bba1d9029738ce4a14a848"
   end
 
-  resource "prompt_toolkit" do
-    url "https://files.pythonhosted.org/packages/source/p/prompt_toolkit/prompt_toolkit-3.0.51.tar.gz"
-    sha256 "ee2b1d8a1de5c6f0a9e0b28c5c7b0e48dc3bedd0c9cbf7c62d79ad27012a41c"
+  resource "ki-protocol" do
+    url "https://files.pythonhosted.org/packages/88/61/0c7927630aa8300d098a7614fa145395aa19a66b32358a9c345a58aa31a7/ki_protocol-1.0.0.tar.gz"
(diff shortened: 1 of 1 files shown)


MSKazemi (approved):
Merging this. Every objectively checkable claim in it holds, and I verified them independently rather than reading the diff — details below, including the one thing neither of us could test.

**First, the thing that matters most about this PR.** You wrote:

> Homebrew itself, Ruby, and WSL are not installed in the development environment, so `brew audit --strict` and `brew install --build-from-source` could not be run locally. I am reporting that limitation explicitly rather than treating the static and PyPI checks as an install test.

That is exactly right, and I am in the same position — **I do not have Homebrew on this machine either**, so I could not run those two commands for you as I offered. I would rather say that plainly than let this merge look better-verified than it is. What I could check, I checked properly.

## What I verified independently

**All 19 sha256 values, by downloading the bytes and hashing them** — not by comparing against PyPI's recorded digest, which would only prove you copied the metadata correctly:

```
kube-q (main)      OK  (247459 bytes)      pydantic           OK  (844775 bytes)
anyio              OK  (260176 bytes)      pydantic-core      OK  (471464 bytes)
annotated-types    OK  (15893 bytes)       pygments           OK  (4955991 bytes)
h11                OK  (101250 bytes)      pyyaml             OK  (130960 bytes)
httpcore           OK  (85484 bytes)       rich               OK  (230680 bytes)
httpx              OK  (141406 bytes)      typing-extensions  OK  (113555 bytes)
idna               OK  (196711 bytes)      typing-inspection  OK  (75949 bytes)
ki-protocol        OK  (4536 bytes)        wcwidth            OK  (1466253 bytes)
markdown-it-py     OK  (82454 bytes)       mdurl              OK  (8729 bytes)
prompt-toolkit     OK  (435492 bytes)
```

19/19. For contrast, the formula this replaces had **two 63-character values** — not stale hashes, structurally invalid ones.

**The dependency closure is exactly complete.** I resolved `kube-q==1.5.0` for Python 3.12 independently and diffed against your resource list:

```
MISSING from formula : none
EXTRA in formula     : none
```

18 resources plus Homebrew's `certifi` covers the full 19-package runtime graph, with nothing carried that is not needed. The old formula had four resources for what is a seven-dependency package.

**Every direct constraint is satisfied** — `rich>=14.3.3` → 15.0.0, `pyyaml>=6.0.3` → 6.0.3, `pygments>=2.18` → 2.20.0, and so on for all seven.

**Both Homebrew constructs I could not test are real**, so I checked them against Homebrew's source rather than guessing:

- `certifi` is a genuine `homebrew/core` formula, currently `2026.7.22` — the exact version the resolver picks. Excluding it from the PyPI resources and taking it from Homebrew is the correct pattern, not a shortcut.
- `pypi_packages exclude_packages: "certifi"` is valid DSL. I checked [`formula.rb`](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/formula.rb), where the signature is `T.any(String, T::Array[String])` and the body does `exclude_packages = Array(exclude_packages)` — so the bare string is idiomatic, not a type error waiting to happen.

**And the three defects from #56 are all fixed:** `license "MIT"` → `AGPL-3.0-or-later`, `homepage` → `MSKazemi/kubeintellect` (the #74/#78 defect, which the issue never mentioned — you fixed it anyway), and the URL now points at 1.5.0 rather than the pre-relicense 1.0.0.

The licence line is the one that actually mattered. A formula telling users AGPL code is MIT is a licensing misstatement, and it is now correct.

## What remains untested, honestly

`brew audit --strict` and `brew install --build-from-source` have **not** been run by anyone. The `maturin`/`rust` build path for `pydantic-core` and the `libyaml` path for PyYAML are reasoned, not observed. I'm merging anyway because the file is not a tap — nobody can `brew install` from a source tree regardless — so the blast radius today is zero, while the licence misstatement was real and live. I've opened a follow-up so this is tracked rather than quietly assumed.

Your judgement call on Option A over Option B was the right one, and you made it after actually working the problem rather than before.

This is your second merged contribution. Thank you — and if KubeIntellect is useful to you, a star helps other engineers find it and passing it on helps more; both entirely optional.

PR #112, opened 2026-08-10
feat(playbooks): add Deployment rollout stuck (ProgressDeadlineExceed…

### LINKED ISSUE #98: playbook: Deployment rollout stuck (ProgressDeadlineExceeded)
The detector engine ships **18 declarative failure playbooks**. This proposes one more: **Deployment rollout stuck (ProgressDeadlineExceeded)**.

## Why this one

A stuck rollout leaves both old and new ReplicaSets alive. The application often still serves traffic, so nothing pages — and the deployment quietly never completes.

## Where it goes

`v4/packages/kubeintellect-server/app/agent/playbooks/rollout_stuck.yaml`

```yaml
name: <PascalCaseName>
detect:
recommended_fix_template: |
```

## What to put in it

**Signals to detect:** `kubectl rollout status` never returning; condition `Progressing=False` with `reason=ProgressDeadlineExceeded`; two ReplicaSets with non-zero replicas

**Notes:** Root cause is usually downstream — the new pods are failing readiness, unschedulable, or stuck pulling an image. The playbook should chain to the existing readiness/pending/imagepull playbooks rather than duplicating them.

## Acceptance criteria

- [ ] The file parses: `python3 -c "import yaml,sys; yaml.safe_load(open('v4/packages/kubeintellect-server/app/agent/playbooks/rollout_stuck.yaml'))"`
- [ ] Both suites still pass

## How to test

```bash
```

## Task metadata

| | |
|---|---|

hariomlohardev:
## Summary

Add a declarative playbook for Deployment rollouts stuck in `ProgressDeadlineExceeded`.

The playbook detects a Deployment with `Progressing=False` and `reason=ProgressDeadlineExceeded`, including Event and PromQL signals. It investigates the rollout using real `kubectl` commands across the Deployment, ReplicaSets, Pods, and Events, then points operators toward the existing readiness, pending, and image-pull playbooks for downstream causes.

The playbook uses a 600-second debounce to avoid firing on transient rollout states and only describes recommended remediation; nothing is automatically applied.

No Python changes — YAML only.

## What & why

Closes #98

A Deployment can remain stuck while the previous ReplicaSet continues serving traffic and the new ReplicaSet fails to become Available. This playbook makes that failure mode explicitly detectable and provides an ordered investigation path without duplicating the existing downstream failure playbooks.

## Type of change

* [ ] 🐛 Bug fix
* [x] ✨ New feature
* [ ] 📖 Docs
* [ ] 🧹 Refactor / chore
* [ ] ⚡ Performance
* [ ] 🧪 Tests

## Scope

* Version directory touched: `v4/`

## Validation

* YAML parses successfully with `yaml.safe_load`
* 19 playbooks detected after the addition
* 19 unique playbook names
* `DeploymentRolloutStuck` is unique and PascalCase
* `debounce_seconds` is set to `600`
* 6 investigation steps are present
* 4 expected-evidence entries are present
* `recommended_fix_template` is present
* `git diff --check` passes
* `kube-q` test suite: 297 passed
* No live Kubernetes reproduction was performed

## Checklist

* [ ] New behavior has both a **happy-path** and an **error-path** test — not applicable; this is a declarative YAML playbook with no Python test added.
* [x] **Every mutating/write operation keeps its dry-run + diff + human-approval (HITL) gate** — no mutating operation is introduced; the recommended fix is explicitly not auto-applied.
* [x] Secret values are never logged or returned (key names only)
* [x] Relevant local tests pass (`kube-q`: 297 passed)
* [ ] `uv run ruff check .` passes locally — not applicable to the YAML-only change
* [ ] `uv run mypy src` passes locally — not applicable to the YAML-only change
* [ ] Docs updated if behavior/CLI/flags changed — no behavior/CLI documentation changes
* [x] Commit is signed off (`git commit -s` — DCO)

## Notes for reviewers

The expected evidence is deliberately phrased to distinguish a genuinely stuck rollout from downstream failures such as `Pending`, `ImagePullBackOff`, or readiness failures.

The playbook does not claim live-cluster evidence; the signals and investigation flow are based on the Kubernetes failure mode described in the issue and the existing playbook schema.


MSKazemi (approved):
Merging this. The playbook itself needed no changes — I only fixed the counts that adding a 20th playbook moves, and pushed that to your branch.

**This is the best-designed playbook in the set, and I want to say why rather than just merge it.**

It **routes instead of duplicating.** `ProgressDeadlineExceeded` is almost never the cause — it is what the Deployment controller says when something else went wrong for ten minutes. You could have written a playbook that re-explains image pulls and probe failures. Instead the investigation steps chain out to `ReadinessProbeFailing`, `PendingInsufficientResources`, `PendingSchedulingConstraints` and `ImagePullBackOff`, and the fix template opens with **"do not force-delete pods."** That is the instruction that saves someone from making it worse.

And this line is new to the library:

> "Absence of ProgressDeadlineExceeded and a single ReplicaSet at desired replicas means the rollout is not stuck — look at ImagePullBackOff, PendingSchedulingConstraints, or ReadinessProbeFailing instead"

**An `expected_evidence` entry that names when the playbook does *not* apply.** None of the other nineteen do that, and it is exactly what stops an agent from confirming the hypothesis it started with. The 600-second debounce is the same instinct — a rollout that is merely slow is not a rollout that is stuck.

You also put `triggers:` at the top level and gave `detect:` real predicates, so it compiles to a detector rather than sitting inert. That was the trap the previous playbook PR fell into; yours avoided it.

## What I changed (counts only, no playbook edits)

Your PR was red, and it was not your fault: `main` gained a 19th playbook (#108) about an hour before you opened this, so the numbers your branch was written against had already moved. Adding a playbook touches **six** numbers across two gates:

- `test_doc_claims.py` — playbook count 19 → 20 in `agent-behaviors.md`, `capabilities.md`, `glossary.md`, `architecture.md`
- `test_detectors.py` — yours *compiles*, so the detector count went 16 → 17 (also in `api-reference.md`). I renamed `test_sixteen_compiled_two_llm_only` → `test_seventeen_compiled_three_llm_only` and added assertions that `DeploymentRolloutStuck` is in the compiled set and `NetworkPolicyBlocking` is not.

Your own validation said "19 playbooks detected" and "`kube-q` test suite: 297 passed" — both correct on your base, and both stale against `main` (20 and 312). Worth knowing that a green run on an older base is not a prediction about CI.

I also added three trigger tests, including one asserting a healthy `ScalingReplicaSet` event does **not** fire it.

## Verification (on the merged tree)

```
server suite   1001 passed        (998 + the 3 added here)
kube-q suite   312 passed
ruff           All checks passed
mypy           Success: no issues found in 171 source files
doc claims     pass
detectors      17 compiled, 3 LLM-only
```

All ten required checks green.

Your commit stays as the feature with mine on top as the count fix, and you're credited in `CHANGELOG.md` and the release notes. That is two merged today — thank you, genuinely.

If the project is useful to you, a star helps other engineers find it and sharing it helps more; both optional as always.

Issue #20, opened 2026-08-02
Execute stored PromQL in playbook detect blocks


MSKazemi:
Playbook `detect:` blocks can carry a `promql` expression, and the detector model supports a PromQL source — but stored **instant** queries in detect blocks aren't executed yet. Closing this loop enables zero-LLM-token, metric-based firings.

### Scope
- Add the execution path in `v4/packages/kubeintellect-server/app/detectors/engine.py`, using the existing `prometheus_tool`.
- Tests.

Medium context in the detector engine.

PR #114, opened 2026-08-11
feat(playbooks): add HPANotScaling playbook for HPA metric failures

### LINKED ISSUE #97: playbook: HorizontalPodAutoscaler not scaling
The detector engine ships **18 declarative failure playbooks**. This proposes one more: **HorizontalPodAutoscaler not scaling**.

## Why this one

An HPA that silently does nothing looks exactly like an HPA that is working and simply not needed — which is why it is usually discovered during an incident rather than before one.

## Where it goes

`v4/packages/kubeintellect-server/app/agent/playbooks/hpa_not_scaling.yaml`

```yaml
name: <PascalCaseName>
detect:
recommended_fix_template: |
```

## What to put in it

**Signals to detect:** `kubectl get hpa` showing `TARGETS=<unknown>/50%`; conditions `AbleToScale=False` or `ScalingActive=False`; events mentioning `failed to get cpu utilization`

**Notes:** The dominant cause is metrics-server missing or unreachable. The second is a container with **no `resources.requests.cpu`**, which makes a utilization target uncomputable.

## Acceptance criteria

- [ ] The file parses: `python3 -c "import yaml,sys; yaml.safe_load(open('v4/packages/kubeintellect-server/app/agent/playbooks/hpa_not_scaling.yaml'))"`
- [ ] Both suites still pass

## How to test

```bash
```

## Task metadata

| | |
|---|---|

Chris7717:
Closes #97

Reproduced both root causes locally against a kind cluster:
- metrics-server missing/unreachable → apiservice v1beta1.metrics.k8s.io not found, HPA condition ScalingActive=False/FailedGetResourceMetric
- container missing resources.requests.cpu → same reason, message "missing request for cpu in container ..."

investigation_steps and expected_evidence are taken from real kubectl output, not invented.
Both suites pass locally (1001 passed); ruff and mypy clean.

**AI assistance disclosure:** used Claude Code to walk through the playbook schema and review each field against my own kind-cluster reproduction; the doc/test count bumps (21/18 playbooks/detectors) were also applied with its help since they're mechanical. I understand the file and can defend each design choice in review.


Chris7717:
Addressed the checklist:

1. All 6 local gates pass (`make setup` from repo root, clean).
2. Added dedicated tests for this change: `test_match_hpa_not_scaling_by_metrics_server_missing`
   and `test_match_hpa_not_scaling_by_missing_cpu_request` in tests/test_playbooks.py, using the
   real `kubectl get events` output captured while reproducing both root causes on a local kind
   cluster (not synthetic data).
3. This PR doesn't touch any diagnosis/report/summary-generating code — it's a declarative
   playbook file plus doc/test count updates — so the "run against an absent resource" check
   doesn't apply here.
4. AI assistance disclosed in the PR description above.

MSKazemi:
Merged — thank you. This is a good playbook, and the reproduction is what made it reviewable.

The part worth calling out: separating the two root causes that Kubernetes reports through the *same* two event reasons but that need completely different fixes — metrics-server missing/unreachable versus a container with no `resources.requests.cpu` — is exactly the discipline this library needs. Transcribing `investigation_steps` and `expected_evidence` from real `kubectl` output rather than reasoning them is what makes a playbook trustworthy, and it is why this one was straightforward to check. You also put `triggers:` at the top level, which is the one schema trap in this directory that produces a silent no-op; two earlier playbook PRs got that wrong.

I pushed one commit on top rather than sending it back. Two changes are worth knowing about, because neither was visible to any gate this repo runs.

**1.** The `detect:` predicate was `"^(FailedGetResourceMetric | FailedComputeMetricsReplicas)$"`. The spaces bind *into* the alternation branches, so the compiled pattern requires an event reason containing a space — and a Kubernetes event reason is a CamelCase identifier, so it could never match anything. The detector still loaded, still counted toward the 18, and still passed the schema check, while being a permanent no-op on the zero-token detection path. Now [anchored without the spaces](https://github.com/MSKazemi/kubeintellect/blob/4ae345ee1e659898a91ab2a5117feeeea2528b19/v4/packages/kubeintellect-server/app/agent/playbooks/hpa_not_scaling.yaml#L4-L6).

**2.** The stored PromQL read `kube_horizontalpoduatoscaler_status_condition` and matched `status="False"`. kube-state-metrics emits [`kube_horizontalpodautoscaler_status_condition`](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/metrics/workload/horizontalpodautoscaler-metrics.md), and the `status` label is lower-case `true`/`false`/`unknown` — the convention `node_not_ready.yaml` and `rollout_stuck.yaml` already follow. That arm is latent, since `engine.py` does not execute stored PromQL yet (#20), but it was stored wrong.

Neither of these was really yours to catch. Your tests exercise `match_playbooks()`, and until this PR **nothing in the suite asserted that a compiled `detect:` predicate can actually fire** — `triggers:` and `detect:` are two independent features, and the playbook and detector counts prove neither of them works. So the fix commit also adds `TestHPANotScalingDetector` (both real reasons, plus a healthy-HPA and a wrong-`involved_kind` negative) and a class guard that rejects any shipped `reason_regex` whose alternatives carry surrounding whitespace. All three fail against the original file; I checked by reverting it.

Your two commits are unchanged, and you are credited in `CHANGELOG.md` and in the release notes.

Verified on `4ae345ee1e659898a91ab2a5117feeeea2528b19`: ruff, mypy (171 files), file modes, syntax warnings, doc-claims, and both suites on 3.12 and 3.13 — 1008 server tests (up from 1003) and 312 CLI. All eleven checks green.

If you want another, #96 (in-cluster DNS resolution failing) is the closest neighbour to this one.

MSKazemi:
Following up on this months later with something I owe you.

I audited the project's contributor recognition today and found you were on **neither** surface
— not in the README Contributors table, not in `.all-contributorsrc`. That was wrong, and it
had been wrong since #114 merged. Both are fixed on `main` now
([#169](https://github.com/MSKazemi/kubeintellect/pull/169),
[`762a26f`](https://github.com/MSKazemi/kubeintellect/commit/762a26f5511074713efeeaa461d2426952d1b3e2)),
listed as `code` and `bug`, with a row in the
[Contributors table](https://github.com/MSKazemi/kubeintellect#contributors) that names both
things you did.

**On the playbook.** `HPANotScaling` is in the shipped set and loads on every server start —
you can see it in the loader output in any CI run. What I wrote in the table is the part that
mattered to me: you reproduced **both** root causes on a real kind cluster (metrics-server
unreachable, and a container with no `resources.requests.cpu`) and took the investigation steps
and expected evidence from actual `kubectl` output. Most playbook contributions are written
from the schema. Yours was written from a cluster, and that is the difference between a
playbook that reads well and one that helps at 3am.

**On #115.** Your report was better than it was given credit for. It pointed straight at the
right file, and chasing it down turned up a cause nobody had suspected: `.gitignore`'s `.env.*`
glob matches `.env.example` too, so the template could not be committed back **at all** — which
is why it had never been restored after the commit you named. That one report is the reason a
`.gitignore` line that had been quietly breaking every fresh clone got found.

You disclosed AI assistance in the PR description without being asked to. That is exactly the
norm the project wants, and you set it before `CONTRIBUTING.md` was as clear about it as it is
now.

Nothing needed from you here — I just did not want the roster fix to happen silently.

PR #116, opened 2026-08-12
fix(memory): make recall similarity floor configurable

### LINKED ISSUE #14: Make the memory recall similarity noise-floor configurable

uuzzrm:
## What & why

Closes #14

The trigram similarity floor used by episodic recall and theme-summary recall was hard-coded independently in both modules. This change adds a validated `MEMORY_RECALL_SIMILARITY_FLOOR` setting with the existing `0.02` default and uses it on both recall paths.

## Type of change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] ⚡ Performance
- [ ] 🔒 Security
- [ ] 📝 Documentation
- [ ] 🧹 Refactor / chore

## Scope

- [x] No behavior change outside the selected version
- [x] No secrets, API keys, or credentials added
- [x] Existing default behavior is preserved

## Checks

- [x] `ruff check packages/kubeintellect-server/app/ packages/ki-protocol/`
- [x] `mypy packages/kubeintellect-server/app packages/ki-protocol packages/kube-q/kube_q`
- [x] Server suite: `1011 passed, 1 warning`
- [x] Kube-q suite: `312 passed`
- [x] `make check-modes`
- [x] `make check-syntax`
- [x] Commit includes DCO sign-off

## Notes for reviewers

The hybrid recall SQL now receives the configured threshold, while baseline episodic and summary post-filters read the same setting. The default remains `0.02`.

AI assistance: I used AI assistance to help inspect the existing recall paths and draft the focused tests and patch. I reviewed the diff and validation output before submitting.

### DIFF
--- a/v4/packages/kubeintellect-server/app/core/config.py
+++ b/v4/packages/kubeintellect-server/app/core/config.py
@@ -210,6 +210,9 @@ class Settings(BaseSettings):
     # baseline single-channel trgm recall. Additive; no schema change required
     # (ts_rank is computed inline; a stored tsvector + GIN index is the P0 speedup).
     MEMORY_HYBRID_RETRIEVAL: bool = False
+    # Minimum pg_trgm similarity used by the baseline recall and summary channels.
+    # Keep this configurable for deployments with different memory vocabularies.
+    MEMORY_RECALL_SIMILARITY_FLOOR: float = Field(default=0.02, ge=0.0, le=1.0)
     # Memory V5 P2 (ADR-013): bi-temporal KG. When on, edge writes stamp event-time
     # valid_from (from the observation) and supersede via retract (retracted_at) instead
     # of only closing (valid_to); point-in-time `as_of(valid_t, tx_t)` queries become
--- a/v4/packages/kubeintellect-server/app/memory/episodes.py
+++ b/v4/packages/kubeintellect-server/app/memory/episodes.py
@@ -25,8 +25,6 @@ _pool = None  # asyncpg.Pool — owned by app.memory.service
 
 # Reciprocal Rank Fusion constant (Cormack et al. 2009). Memory V5 ADR-014 / P1.
 _RRF_K = 60
-# Baseline trgm noise floor — unrelated episodes below this don't help the prompt.
-_TRGM_FLOOR = 0.02
 
 # ── Memory V5 P6 (ADR-017): importance / surprise ────────────────────────────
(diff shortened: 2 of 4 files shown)


MSKazemi:
Merged via #129 — **unchanged**. Thank you, this was a genuinely clean PR.

The thing I want to name, because it is the part that is easy to skip: you did not just hoist the
constant, you noticed that the hybrid RRF path must *not* re-apply the floor post-fetch and left
that asymmetry intact. A lexical-only match legitimately has `sim <= floor`, so re-filtering it
would have silently deleted exactly the rows ADR-014 added the second channel to find. That is the
kind of detail a refactor usually eats, and the comment you left above it means the next person
will not undo it either.

The two tests are the right two: one pins the default so the setting cannot drift from `0.02` by
accident, and the other patches the setting and proves the override actually reaches both recall
paths — rather than asserting the constant equals itself, which is the usual shape of this test and
proves nothing.

Verified on the merge result, not just your branch: `ruff` clean, `mypy` 0 errors across 171 files,
server suite **1020 passed**, `kq` suite **312 passed**.

I have added `code` to your entry in `.all-contributorsrc` and extended your line in the README's
contributors table.

---

One optional thing, entirely your call and it changes nothing here either way: if you would like to
post about this — LinkedIn, X, Mastodon, a Kubernetes community, wherever you are active — please
do. It is your work and the credit is yours; a configurable recall floor in an open-source AI SRE
is a nice concrete thing to point at. If you tag the repo I will happily amplify it from the
project's side. And if you would rather not, that is completely fine — no expectation attached.

If you want a next one: #20 is open and adjacent to what you have already read — the detector
engine stores PromQL in playbooks but never executes it, so the `promql:` blocks are currently
documentation. You now know that subsystem better than most.

PR #117, opened 2026-08-12
fix: restore v4/.env.example (#115)

### LINKED ISSUE #115: bug: .env.example missing — dropped in publish/sync commit 96ae100, never restored for v4
### Version / generation

v4 @ commit 96ae1006 (also affects v1–v3, see note below)

### Component

Deploy / Helm / infra

### What happened?

1. I cloned the repo and followed the v4 setup docs, which say `cp .env.example .env`.
2. Expected: a `.env.example` template to exist somewhere in the repo (root or v4/).
3. Actual: `.env.example` does not exist anywhere in the tree, even though docs/README/Makefile
   for v4 (and v1–v3) still reference it.

Root cause: commit 96ae100 ("publish: sync public mirror to current KubeIntellect source (v1–v4)")
restructured the repo from a flat layout into v1/ v2/ v3/ v4/. In that same commit it deleted the
root .env.example (244 lines) and added `.env` / `.env.*` to .gitignore — but never recreated an
.env.example for the new v4/ layout. The .gitignore change is not the cause; the file was already
gone from that commit's diff. v1–v3 have the same gap, but per TRIAGE.md those are frozen, so this
report is scoped to fixing it for v4 only.

### Steps to reproduce

1. git clone <repo>
2. find . -iname ".env.example"   → no results
3. Try to follow any "cp .env.example .env" instruction in v4/README.md, v4/docs/deploy/*.md,
   or v4/Makefile → the file to copy does not exist.

### Relevant logs / output

```shell
git show --stat 96ae1006fe0596a1da44084d0a3f7c0d77ca2712 | grep -i env
 .env.example                                       |  244 -
 .gitignore                                         |   22 +

git log --all --follow --oneline -- .env.example
96ae100 publish: sync public mirror to current KubeIntellect source (v1–v4)
1b1f9a2 Enhance authentication system and update documentation
1369943 Enhance environment configuration and deployment documentation
66d8d66 Refactor environment configuration and remove deprecated files
47fd8ed Add initial project structure and documentation
```

### Environment

N/A — this is a repo/packaging issue, reproducible on any OS straight after clone.

### Before submitting

- [x] I searched existing issues and this isn't a duplicate.
- [x] I redacted secrets/tokens from any logs above.

hariomlohardev:
## What & why

Closes #115

Allow .env.example via .gitignore negation and restore v4 template from 96ae100^ (244 lines, placeholders only).

## Type of change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📖 Docs
- [ ] 🧹 Refactor / chore
- [ ] ⚡ Performance
- [ ] 🧪 Tests

## Scope

## Checklist

- [ ] New behavior has both a **happy-path** and an **error-path** test

## Notes for reviewers

No real secrets — `v4/.env.example` is placeholder-only (`POSTGRES_PASSWORD=changeme`, `AZURE_OPENAI_API_KEY=` empty). `.gitignore` now allows `!.env.example` / `!**/.env.example` after `.env.*`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,10 @@
 # ── Secrets ──────────────────────────────────────────────────────────────────
 .env
 .env.*
+# Allow the example template to be versioned — it contains only placeholders.
+# Must come *after* .env.* so the negation takes effect (git check-ignore is last-match-wins).
+!.env.example
+!**/.env.example
 *.local
 *.pem
 *.key
--- a/v4/.env.example
+++ b/v4/.env.example
@@ -1,0 +1,244 @@
+# KubeIntellect — Runtime Configuration
+#
+# Usage:
+#   cp .env.example .env
+#   # Edit .env — fill in at minimum: LLM key + POSTGRES_PASSWORD + one ADMIN key
+#   docker compose up -d
+#
+# This file is read by Docker Compose only.
+# Kubernetes (Helm) deployments use a ConfigMap + Secret — not this file.
+#
+# NEVER commit .env to git — it is already listed in .gitignore.
+# Generate secure keys with: openssl rand -hex 20
+#
+# Full variable reference: https://mskazemi.github.io/kubeintellect/configuration/
+
+
+# ══════════════════════════════════════════════════════════════════════════════
+# LLM PROVIDER  (required — pick one)
+# ══════════════════════════════════════════════════════════════════════════════
+#
+# Which LLM backend to use. Must be exactly: azure  or  openai
+
+LLM_PROVIDER=azure
+
+
+# ── Option A: Azure OpenAI ────────────────────────────────────────────────────
+#
+# Required when LLM_PROVIDER=azure.
+# Find these in Azure Portal → your OpenAI resource → Keys and Endpoint.
+
+AZURE_OPENAI_API_KEY=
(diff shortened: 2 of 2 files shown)


MSKazemi:
Merged via #129 — thank you.

The part worth calling out: the issue blamed a commit for deleting the file, and you did not stop
there. You found the *live* cause — `.gitignore`'s `.env.*` matches `.env.example`, so the file
could not be re-added at all until that pattern was negated. Restoring the file without the
negation would have looked like a fix and quietly done nothing. Your comment on the negation lines
explaining last-match-wins is exactly right and I kept it verbatim.

**One change on merge.** You restored the template as it was before it was lost, and it has aged:
the 244-line version predates the `qwen` and `anthropic` providers (both supported now, and both
asserted by `test_doc_claims`) and still tells people to copy Langfuse keys out of the UI, which
`make langfuse-provision` replaced. I merged the current 289-line template instead, so the file
matches what the code actually reads. No way you could have known that from the repo — the newer
version only existed on disk, because the same `.gitignore` line was hiding it.

Scanned before publishing, since this file is now public: every credential field is empty or an
obvious placeholder (`changeme`, `langfuse-local`), no real values.

PR #118, opened 2026-08-12
fix: brew formula style for audit --strict (#113)

### LINKED ISSUE #113: Run `brew audit --strict` and a real `brew install` against the updated kube-q formula
`Formula/kube-q.rb` was corrected in #111 (licence, homepage, version 1.5.0, and the full 18-resource dependency tree). **Every static check passed, but nobody has actually installed it.**

## What is already verified — do not redo this

- **All 19 sha256 values** were verified by downloading the bytes and hashing them (not by re-reading PyPI metadata). 19/19 match.
- **The dependency closure is exactly complete**: resolving `kube-q==1.5.0` for Python 3.12 gives 19 runtime packages; the formula carries 18 resources plus Homebrew's `certifi`, with nothing missing and nothing extra.
- **All seven direct constraints** from `kube-q` 1.5.0 are satisfied by the pinned versions.
- `certifi` is a real `homebrew/core` formula, and `pypi_packages exclude_packages: "certifi"` is valid DSL (`formula.rb` coerces with `Array()`).

## What is genuinely unverified

Neither the contributor nor the maintainer has Homebrew available, so **these two commands have never been run**:

```bash
brew audit --strict --online Formula/kube-q.rb
brew install --build-from-source ./Formula/kube-q.rb
kq --version        # expect 1.5.0
```

Specifically unproven:

1. **`pydantic-core` builds from sdist** — it needs the `maturin` and `rust` build dependencies the formula declares. This is the most likely thing to fail, and it is a compile step, not a download.
2. **PyYAML builds against Homebrew's `libyaml`.**
3. **`brew audit --strict` style compliance** — ordering, `desc` rules, and whether `depends_on "certifi" => :no_linkage` audits cleanly.

## What to report

Paste the real output either way. **A failure report is worth as much as a success** — more, if it names the step that broke. If `brew audit` emits style warnings, list them rather than fixing them silently, so we can decide which are worth acting on.

If it installs cleanly, say so and this closes.

## Why this is not urgent

A formula sitting in a source tree is not a tap — nobody can `brew install` from here regardless, so nothing is broken for users today. This is about whether Homebrew becomes a channel we can honestly claim. **If it turns out to need ongoing upkeep nobody will do, "delete the formula" remains a legitimate outcome** and would close this too.

---

| | |
|---|---|
| **Difficulty** | Beginner — but needs macOS or Linuxbrew |
| **Estimated time** | 30 minutes, most of it waiting for a Rust compile |
| **Skills** | A working `brew`. No Python, no cluster, no API key. |
| **Mentor** | @MSKazemi — ask on this issue |
| **Status** | 🟢 **Unclaimed** — comment "I'd like to take this" and it will be labelled `claimed` |

Follows #56 / #111.

hariomlohardev:
## What & why

Closes #113

Sort resources alphabetically and shorten desc to 75 chars; add verify-brew.sh for static + brew audit/install checks.

## Type of change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📖 Docs
- [ ] 🧹 Refactor / chore
- [ ] ⚡ Performance
- [ ] 🧪 Tests

## Scope

- Version directory touched: `v4/packages/kube-q/`

## Checklist

- [ ] New behavior has both a **happy-path** and an **error-path** test

## Notes for reviewers

Static `bash v4/packages/kube-q/scripts/verify-brew.sh` → PASS (desc ≤80, resources sorted, sha256 format). Real `brew audit --strict --online` + `brew install --build-from-source && kq --version` requires macOS/Linuxbrew with `rust`+`maturin`+`libyaml` — paste output when run.

### DIFF
--- a/v4/packages/kube-q/Formula/kube-q.rb
+++ b/v4/packages/kube-q/Formula/kube-q.rb
@@ -1,7 +1,7 @@
 class KubeQ < Formula
   include Language::Python::Virtualenv
 
-  desc "Interactive terminal CLI for chatting with your Kubernetes cluster via an AI backend"
+  desc "Interactive CLI for chatting with your Kubernetes cluster via an AI backend"
   homepage "https://github.com/MSKazemi/kubeintellect"
   url "https://files.pythonhosted.org/packages/92/85/5d0af5d4fcff49447325ffd338c2ef5e386d99b6a106b570f8e97bb66d60/kube_q-1.5.0.tar.gz"
   sha256 "738363d6f3c5c1f97c561924ab232e2cfc0ca6cacb5c9bdb18ab756aa1b1cfc1"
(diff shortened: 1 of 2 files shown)


MSKazemi:
Merged via #129 — thank you, and thank you especially for `verify-brew.sh`. #113 had been sitting
on two commands nobody could run; making the static half reproducible with nothing but `bash` is
the right way to unstick that, and it will outlive the issue.

**Two corrections on merge**, and the first one is a nice trap:

`brew audit` measures **`"<name>: <desc>"`**, not the description on its own. Your shortened desc
is 75 characters, which is comfortably under 80 — but with the `kube-q: ` prefix it is 83, so
`brew audit --strict` would still have flagged it. And because the script asserted `desc <= 80`,
it reported `PASS` on a formula brew rejects. Both now use the real rule; the description is
78 including the prefix.

That is the same shape as the bug in #113's own re-verification, and it is worth internalising: a
check calibrated to a slightly different rule than the one that matters is more dangerous than no
check, because it converts "unverified" into "verified" without anyone noticing.

Also dropped the unused `ALT_FORMULA` variable, and trimmed the closing lines that advised how to
triage the issue — the script is better as a pure verifier.

`brew audit` and `brew install` themselves are still unrun; there is no Homebrew on any machine
here either, so I have left #113 open for that specific step rather than closing it on static
checks. Your script is what someone with a Mac will run when they get to it.

PR #127, opened 2026-08-12
fix: add PVC Pending playbook (#94)

### LINKED ISSUE #94: playbook: PersistentVolumeClaim stuck Pending
The detector engine ships **18 declarative failure playbooks**. This proposes one more: **PersistentVolumeClaim stuck Pending**.

## Why this one

A PVC that never binds silently blocks every pod that mounts it. The pod sits `Pending` with a message about the volume, which sends people looking at scheduling rather than storage.

## Where it goes

`v4/packages/kubeintellect-server/app/agent/playbooks/pvc_pending.yaml`

```yaml
name: <PascalCaseName>
detect:
recommended_fix_template: |
```

## What to put in it

**Signals to detect:** `kubectl get pvc -A` showing `STATUS=Pending`; `kubectl describe pvc` events such as `no persistent volumes available for this claim` or `storageclass.storage.k8s.io "x" not found`

**Notes:** The most common real causes are a missing/misspelled `storageClassName`, no default StorageClass in the cluster, or a static PV whose `accessModes`/capacity do not match the claim.

## Acceptance criteria

- [ ] The file parses: `python3 -c "import yaml,sys; yaml.safe_load(open('v4/packages/kubeintellect-server/app/agent/playbooks/pvc_pending.yaml'))"`
- [ ] Both suites still pass

## How to test

```bash
```

## Task metadata

| | |
|---|---|

hariomlohardev:
## What & why

Closes #94

Add declarative playbook `PvcPending` for PVC stuck `Pending` (missing StorageClass, no default class, PV mismatch). No Python — `detect → investigate → remediate` YAML.

## Type of change

- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 📖 Docs
- [ ] 🧹 Refactor / chore
- [ ] ⚡ Performance
- [ ] 🧪 Tests

## Scope

## Checklist

- [x] New behavior has both a **happy-path** and an **error-path** test — `python3 -c "import yaml; yaml.safe_load(open('v4/.../pvc_pending.yaml'))"` PASS + `pytest` doc-claims
- [x] `uv run pytest` passes locally — 1008 passed (updated `22 playbooks / 19 detectors` in `glossary.md/capabilities.md/architecture.md/agent-behaviors.md/api-reference.md` + `assert len(detectors)==19`)
- [ ] Docs updated if behavior/CLI/flags changed — playbook counts updated where referenced

## Notes for reviewers

Playbook at `v4/packages/kubeintellect-server/app/agent/playbooks/pvc_pending.yaml` (`PvcPending`, `watch_predicates: PersistentVolumeClaim FailedBinding`, `promql: kube_persistentvolumeclaim_status_phase Pending`, `debounce_seconds: 300`). Investigation steps are real `kubectl get/describe pvc|storageclass|pv` in human order. Followed `service_no_endpoints.yaml` schema. `yaml.safe_load` PASS, `uv run pytest v4/tests/test_doc_claims.py` PASS.

### DIFF
--- a/v4/docs/agent-behaviors.md
+++ b/v4/docs/agent-behaviors.md
@@ -123,7 +123,7 @@ detects a matching pattern in the snapshot, the coordinator's system prompt
 includes the playbook(s) inline — guiding it to follow proven steps before
 improvising.
 
-**Playbooks shipped (21):**
+**Playbooks shipped (22):**
 
 *Pod / container lifecycle*
 
@@ -421,7 +421,7 @@ triggers:
 - `detect: null` marks a playbook as **LLM-only** (no machine signal exists,
   or the signal is owned by another playbook).
 
-Of the 21 shipped playbooks, **18 compile to detectors**; 3 are LLM-only
+Of the 22 shipped playbooks, **19 compile to detectors**; 3 are LLM-only
 (`CommandHardcodedFailure` — disambiguated from CrashLoopBackOff only by
 reading the pod spec — `ServiceUnreachable`, and `NetworkPolicyBlocking`,
 where the packet is discarded in the CNI datapath so no machine signal
--- a/v4/docs/api-reference.md
+++ b/v4/docs/api-reference.md
@@ -231,7 +231,7 @@ Response:
 ```json
 {
   "sensorium": "active",
-  "detectors": 18,
+  "detectors": 19,
   "findings": [
     {
       "type": "finding",
--- a/v4/docs/architecture.md
+++ b/v4/docs/architecture.md
@@ -473,7 +473,7 @@ packages/
 │   │   ├── coordinator.py        — LLM decision + synthesis, tool trimmer, reflexion writes
 │   │   ├── memory_loader.py      — load pinned cluster context + failure-pattern hints
 │   │   └── subagent.py           — 4 specialist subagent runners
-│   ├── playbooks/                — YAML playbook library + loader (21 playbooks)
+│   ├── playbooks/                — YAML playbook library + loader (22 playbooks)
 │   ├── state.py                  — AgentState, SubagentInput, AgentFinding, RCAResult
 │   ├── workflow.py               — LangGraph graph, targeted_investigator, route_coordinator
 │   └── hitl.py                   — HITL approval/denial detection
--- a/v4/docs/capabilities.md
+++ b/v4/docs/capabilities.md
@@ -36,7 +36,7 @@ Metrics and logs are optional — without them, KubeIntellect still answers ever
 
 V4 also works *between* your questions:
 
-- **Zero-token known-failure detection** — the playbook library covers the 21 most common failures, and compiled detectors recognize 18 of them straight off the live cluster stream without spending a single LLM token. → [Agent Behaviors → V4 additions](agent-behaviors.md#v4-additions)
+- **Zero-token known-failure detection** — the playbook library covers the 22 most common failures, and compiled detectors recognize 19 of them straight off the live cluster stream without spending a single LLM token. → [Agent Behaviors → V4 additions](agent-behaviors.md#v4-additions)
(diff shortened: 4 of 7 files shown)


MSKazemi:
Merged via #129 — thank you. The investigation steps here are the best part: they walk PVC →
StorageClass → PV in the order a human actually debugs it, and `expected_evidence` naming
*"Pending for >5m (not transient)"* is the kind of line that stops a playbook firing on a cluster
that is simply still starting up.

**One real defect, and it is a subtle one I want to explain properly**, because it has now bitten
this repo twice and I would rather it stop being a trap than be a thing you got caught by.

The `detect:` block shipped as:

```yaml
watch_predicates:
  - kind: PersistentVolumeClaim
    reason_regex: "^(FailedBinding|ProvisioningFailed|StorageClassNotFound)$"
```

`kind:` does **not** name the Kubernetes object — it selects the *observation channel* the
sensorium produces, and there are exactly three: `Pod`, `Event`, `Node`. `WatchPredicate.matches()`
compares against those three and falls through to `return False` for anything else. So this
predicate could never match any observation, ever.

What makes it nasty is that nothing complained. The file parsed, the playbook loaded into the
registry, it counted toward the detector total, it passed `test_each_playbook_has_complete_schema`,
and the whole suite stayed green — while the zero-token detection that #94 actually asked for was
entirely absent. `triggers:` (the prompt-side path) worked fine, which is why the playbook still
*looked* like it was doing something.

The fix is to select the Event channel and narrow it by subject:

```yaml
- kind: Event
  reason_regex: "^(FailedBinding|ProvisioningFailed)$"
  involved_kind: PersistentVolumeClaim
```

I also dropped `StorageClassNotFound`: no controller emits that reason. The missing-StorageClass
case you correctly identified arrives as `ProvisioningFailed` carrying the
`storageclass.storage.k8s.io "x" not found` **message**, which the corrected predicate does catch —
your `triggers:` message regex had it right.

**Guards added so this cannot recur.** `test_every_watch_predicate_uses_a_known_observation_kind`
is now a class guard over every shipped detector, and it names the fix in its failure message.
`TestPvcPendingDetector` proves the predicate fires on both real controller reasons and does not
fire on a success event or on the same reason from another object. All of them fail against the
original file — I checked by reverting.

This is the second time this family has cost us: [#114](https://github.com/MSKazemi/kubeintellect/pull/114)
shipped a detector whose regex had a stray space inside an alternation, equally dead, equally
green. Both times the root cause was the same — nothing asserted that a compiled predicate can
actually *fire*. The counts and the schema check never could. **The takeaway is a repo one, not a
you one: `triggers:` and `detect:` are two independent features, and a playbook PR needs a test
that the `detect:` predicate matches a realistic event.** I am going to put that in the playbook
issue template so the next person is told up front rather than finding out in review.

Counts: this and #128 were stacked, so between them the numbers moved once for two playbooks. The
library is now **23 playbooks / 20 compiled / 3 LLM-only**, and `agent-behaviors.md`'s bullet list —
which neither PR touched — now actually lists 23.

MSKazemi:
Following up on my note above: I said I would put the "prove the `detect:` predicate fires" requirement in the playbook issue template. There is no playbook issue template — #94/#95 were hand-written — so it landed in the two places an author will actually hit instead, in #130:

- **`.github/PULL_REQUEST_TEMPLATE.md`** — a playbook-specific checklist item, so it is in front of you while you open the PR.
- **`v4/docs/agent-behaviors.md`**, immediately under the schema — a warning box spelling out that `kind:` is the observation channel (`Pod` / `Event` / `Node`) and never the Kubernetes object, that `involved_kind:` is how you narrow the subject, and that a shared reason like `Unhealthy` needs a `message_regex` co-condition.

Both point at the tests from your PR as the worked example of the right shape. Sorry for the imprecision in the original comment.

PR #126, opened 2026-08-12
fix: add kq completion worked example (#93)

hariomlohardev:
## What & why

Closes #93

Add worked example for `kq completion` to `v4/docs/examples.md` §10 — full `kq completion bash` script (real output, 1606 chars) + per-shell enable snippets.

## Type of change

- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [x] 📖 Docs
- [ ] 🧹 Refactor / chore
- [ ] ⚡ Performance
- [ ] 🧪 Tests

## Scope

- Version directory touched: `v4/`
- [x] This change is scoped to a version whose contributions are open (`v4/`, or docs/typos in older versions).

## Checklist

- [ ] New behavior has both a **happy-path** and an **error-path** test
- [ ] **Every mutating/write operation keeps its dry-run + diff + human-approval (HITL) gate** (safety invariant)
- [x] Secret values are never logged or returned (key names only)
- [x] `uv run pytest` passes locally — 1008 passed, 27 kube-q tests
- [x] `uv run ruff check .` passes locally
- [ ] `uv run mypy src` passes locally
- [x] Docs updated if behavior/CLI/flags changed

## Notes for reviewers

Verified `kq completion bash` byte-for-byte, headings 1-10 sequential, `grep -c 'kq completion' =9`, markdown balanced, Related links intact. See `cli-reference.md#kq-completion-bashzshfish`.

PR #110, opened 2026-08-10
chore(server,kube-q): modernize datetime.UTC and TimeoutError aliases…

hariomlohardev:
…, sort `__all__`

Ruff 0.16 flags `UP017` (`datetime.timezone.utc` → `datetime.UTC`), `UP041` (`asyncio.TimeoutError` → `TimeoutError`), and `RUF022` (unsorted `__all__`). These are no-ops on Python ≥3.12.

Applied the targeted autofix and removed the four now-unused `datetime.timezone` imports. No behavior was changed.

`UP017`, `UP041`, and `RUF022` now report zero findings.

Fixes #79

## What & why

Modernize the three Ruff 0.16 rule families without changing runtime behavior or touching unrelated Ruff rules.

The deliberate `# noqa: UP017` in `kube_q/cli/store.py` was left unchanged.

## Type of change

* [ ] 🐛 Bug fix
* [ ] ✨ New feature
* [ ] 📖 Docs
* [x] 🧹 Refactor / chore
* [ ] ⚡ Performance
* [ ] 🧪 Tests

## Scope

* Version directory touched: `v4/`
* [x] This change is scoped to a version whose contributions are open (`v4/`, or docs/typos in older versions).

## Verification

* [x] `uv run pytest` passes locally

  * kube-q: **297 passed**
  * server: **986 passed, 2 warnings**
* [x] Targeted Ruff rules pass:

  * `uv run ruff check . --select UP017,UP041,RUF022 --statistics`
  * **0 findings**
* [x] Commits are signed off (`git commit -s` — DCO)

## Checklist

* [ ] New behavior has both a **happy-path** and an **error-path** test
* [x] **Every mutating/write operation keeps its dry-run + diff + human-approval (HITL) gate** (safety invariant) — no mutating operations were changed
* [x] Secret values are never logged or returned (key names only) — no secret-handling code was changed
* [x] `uv run pytest` passes locally
* [x] Targeted `uv run ruff check . --select UP017,UP041,RUF022` passes locally
* [ ] `uv run mypy src` passes locally — not applicable; this repository has no `src/` directory
* [x] Docs updated if behavior/CLI/flags changed — no behavior, CLI, or flags changed
* [x] Commits are signed off (`git commit -s` — DCO)

## Notes for reviewers

This is intentionally a narrow lint modernization.

The changes are limited to:

* `datetime.timezone.utc` → `datetime.UTC`
* `asyncio.TimeoutError` → `TimeoutError`
* sorting `__all__` definitions
[...]

PR #70, opened 2026-08-08
chore(lint): clear EXE002 findings

hariomlohardev:
## What & why

Closes #64

Clears the Ruff `EXE002` findings introduced by executable file mode bits on source files.

* Removes the stray executable (`+x`) bit from **94 source files**.
* Changes are **file-mode only**: `100755 → 100644`.
* No Python source code or test behavior was changed.
* This PR intentionally does **not** address the remaining Ruff 0.16 findings or change the Ruff version pin.

## Type of change

* [ ] 🐛 Bug fix
* [ ] ✨ New feature
* [ ] 📖 Docs
* [x] 🧹 Refactor / chore
* [ ] ⚡ Performance
* [ ] 🧪 Tests

## Scope

* Version directory touched: `v4/`
* [x] This change is scoped to a version whose contributions are open (`v4/`, or docs/typos in older versions).

## Checklist

* [ ] New behavior has both a **happy-path** and an **error-path** test
* [x] **Every mutating/write operation keeps its dry-run + diff + human-approval (HITL) gate** (safety invariant)
* [x] Secret values are never logged or returned (key names only)
* [x] `uv run pytest` passes locally
* [x] `uv run ruff check .` passes locally
* [ ] `uv run mypy src` passes locally
* [ ] Docs updated if behavior/CLI/flags changed
* [x] Commits are signed off (`git commit -s` — DCO)

## Notes for reviewers

Verification:

* Ruff 0.16.1 `EXE002`: **fully cleared**.
* Ruff findings: **436 → 342**, with the reduction entirely attributable to the 94 `EXE002` findings.
* `git diff --stat`: **94 files changed, 0 insertions(+), 0 deletions(-)**.
* All changes are pure file-mode changes: **`100755 → 100644`**.
* Server test suite: **990 passed**.
* kube-q test suite: **312 passed**.
* No source-code or test changes were made.
* The Ruff version pin and Dependabot configuration are intentionally unchanged; the remaining Ruff 0.16 rule families will be handled separately.
* The `mypy src` checklist command is not applicable because this repository does not contain a `src/` directory.

PR #128, opened 2026-08-12
fix: add liveness probe failing playbook (#95)

### LINKED ISSUE #95: playbook: liveness probe failing (repeated restarts)
The detector engine ships **18 declarative failure playbooks**. This proposes one more: **Liveness probe failing (repeated restarts)**.

## Why this one

`readiness_probe_failing.yaml` exists; liveness does not. They present completely differently: a failing readiness probe removes a pod from Service endpoints, while a failing liveness probe **restarts the container**, so the symptom looks like a crash loop with a healthy-looking application.

## Where it goes

`v4/packages/kubeintellect-server/app/agent/playbooks/liveness_probe_failing.yaml`

```yaml
name: <PascalCaseName>
detect:
recommended_fix_template: |
```

## What to put in it

**Signals to detect:** Events with `reason=Unhealthy` and `Liveness probe failed`; a rising `RESTARTS` count with `Running` status

**Notes:** Distinguish it from `crashloopbackoff.yaml`, which is about the process exiting on its own. Here the kubelet is killing a process that is alive but not answering — usually too short an `initialDelaySeconds`/`timeoutSeconds` for a slow-starting app.

## Acceptance criteria

- [ ] The file parses: `python3 -c "import yaml,sys; yaml.safe_load(open('v4/packages/kubeintellect-server/app/agent/playbooks/liveness_probe_failing.yaml'))"`
- [ ] Both suites still pass

## How to test

```bash
```

## Task metadata

| | |
|---|---|

hariomlohardev:
## What & why

Closes #95

Add playbook `LivenessProbeFailing` — distinguishes liveness restarts (kubelet killing a live container) from `CrashLoopBackOff` (process exits on its own). Usually `initialDelaySeconds`/`timeoutSeconds` too short.

## Type of change

- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 📖 Docs
- [ ] 🧹 Refactor / chore
- [ ] ⚡ Performance
- [ ] 🧪 Tests

## Scope

## Checklist

- [x] New behavior has both a **happy-path** and an **error-path** test — `yaml.safe_load` PASS + doc-claims
- [x] `uv run pytest` passes locally — 1008 passed (22 playbooks / 19 detectors docs + test updated)

## Notes for reviewers

`v4/.../liveness_probe_failing.yaml` (`LivenessProbeFailing`, `Event Unhealthy + Liveness probe failed`, `increase(restarts[5m])>2`, `debounce_seconds: 180`). Steps: `kubectl get pods RESTARTS`, `describe pod` (probe spec), `logs --previous`, `get events Unhealthy`. Distinct from `crashloopbackoff.yaml`.

### DIFF
--- a/v4/docs/agent-behaviors.md
+++ b/v4/docs/agent-behaviors.md
@@ -123,7 +123,7 @@ detects a matching pattern in the snapshot, the coordinator's system prompt
 includes the playbook(s) inline — guiding it to follow proven steps before
 improvising.
 
-**Playbooks shipped (21):**
+**Playbooks shipped (22):**
 
 *Pod / container lifecycle*
 
@@ -421,7 +421,7 @@ triggers:
 - `detect: null` marks a playbook as **LLM-only** (no machine signal exists,
   or the signal is owned by another playbook).
 
-Of the 21 shipped playbooks, **18 compile to detectors**; 3 are LLM-only
+Of the 22 shipped playbooks, **19 compile to detectors**; 3 are LLM-only
 (`CommandHardcodedFailure` — disambiguated from CrashLoopBackOff only by
 reading the pod spec — `ServiceUnreachable`, and `NetworkPolicyBlocking`,
 where the packet is discarded in the CNI datapath so no machine signal
--- a/v4/docs/api-reference.md
+++ b/v4/docs/api-reference.md
@@ -231,7 +231,7 @@ Response:
 ```json
 {
   "sensorium": "active",
-  "detectors": 18,
+  "detectors": 19,
   "findings": [
     {
       "type": "finding",
--- a/v4/docs/architecture.md
+++ b/v4/docs/architecture.md
@@ -473,7 +473,7 @@ packages/
 │   │   ├── coordinator.py        — LLM decision + synthesis, tool trimmer, reflexion writes
 │   │   ├── memory_loader.py      — load pinned cluster context + failure-pattern hints
 │   │   └── subagent.py           — 4 specialist subagent runners
-│   ├── playbooks/                — YAML playbook library + loader (21 playbooks)
+│   ├── playbooks/                — YAML playbook library + loader (22 playbooks)
 │   ├── state.py                  — AgentState, SubagentInput, AgentFinding, RCAResult
 │   ├── workflow.py               — LangGraph graph, targeted_investigator, route_coordinator
 │   └── hitl.py                   — HITL approval/denial detection
--- a/v4/docs/capabilities.md
+++ b/v4/docs/capabilities.md
@@ -36,7 +36,7 @@ Metrics and logs are optional — without them, KubeIntellect still answers ever
 
 V4 also works *between* your questions:
 
-- **Zero-token known-failure detection** — the playbook library covers the 21 most common failures, and compiled detectors recognize 18 of them straight off the live cluster stream without spending a single LLM token. → [Agent Behaviors → V4 additions](agent-behaviors.md#v4-additions)
+- **Zero-token known-failure detection** — the playbook library covers the 22 most common failures, and compiled detectors recognize 19 of them straight off the live cluster stream without spending a single LLM token. → [Agent Behaviors → V4 additions](agent-behaviors.md#v4-additions)
 - **Self-opened investigations** — a firing detector opens its own investigation and publishes the report; how far it may go is set per namespace (A0–A3). → [Autonomous Operations](autonomy.md)
 - **Morning digest** — `kq digest` summarizes findings, autonomous investigations, and rollback points from the last N hours. → [Autonomous Operations → digest](autonomy.md#the-morning-digest)
 - **Tamper-evident audit replay** — every session is hash-chained in an append-only log; `kq replay <session-id>` reproduces it and verifies integrity. → [Flight Recorder & Replay](flight-recorder.md)
@@ -48,7 +48,7 @@ V4 also works *between* your questions:
 
 ## Failure patterns it recognizes
 
-KubeIntellect ships **21 built-in playbooks** — deterministic investigation
+KubeIntellect ships **22 built-in playbooks** — deterministic investigation
 recipes for the most common Kubernetes failures. When the cluster snapshot
 matches a pattern, the matching playbook guides the investigation automatically.
 You don't invoke these by name; they fire on their own.
--- a/v4/docs/glossary.md
+++ b/v4/docs/glossary.md
@@ -18,7 +18,7 @@ page where the concept is explained in depth.
 | **Memory loader** | The step that loads pinned, cross-session context (user prefs, past root causes, failure hints) into the prompt. Active only on PostgreSQL. |
 | **Cluster snapshot** | The pre-fetched picture of cluster health (pod list + Warning events) built at the start of every turn. Drives playbook matching and the snapshot sufficiency gate. |
 | **Snapshot sufficiency gate** | A soft bias that lets the agent answer healthy, list-shaped questions straight from the snapshot instead of re-querying. Modes: `off` / `lenient` / `strict`. See [Agent Behaviors](agent-behaviors.md#snapshot-sufficiency-gate). |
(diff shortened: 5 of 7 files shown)


MSKazemi:
Merged via #129 — thank you. You got the hard part of this one right: the `--previous` log step
with the explicit *"was the app healthy before the kubelet killed it, vs CrashLoopBackOff where it
exits on its own"* is precisely the distinction #95 asked for, and the fix template says *"this is
distinct from CrashLoopBackOff"* rather than leaving the operator to work it out. That framing is
what makes the playbook worth having.

**One defect, in the detect block.** It shipped as:

```yaml
- kind: Event
  reason_regex: "^Unhealthy$"
```

The kubelet emits `Unhealthy` for **both** probe kinds — the reason is identical whether it is a
readiness or a liveness probe that failed. Only the message distinguishes them. So this predicate
fired on every readiness failure too, which meant two things at once: it duplicated the existing
`ReadinessProbeFailing` detector on a single event, and it told the operator a container was being
restarted when nothing was being restarted at all. In an incident tool that second one is the
expensive failure — it sends someone to look at probe timings and restart counts for a pod that is
simply out of its Service endpoints.

Fixed by adding the co-condition that names the probe:

```yaml
- kind: Event
  reason_regex: "^Unhealthy$"
  message_regex: "Liveness probe failed"
  involved_kind: Pod
```

**And a change on our side, not yours, that your PR exposed.** `readiness_probe_failing.yaml`
already claimed `"Readiness probe failed|Liveness probe failed"` — it was covering both because
there was no liveness playbook. Now that yours exists, leaving that alternation in place would
double-fire both playbooks on every liveness event, so I narrowed the readiness one to readiness
only and rewrote its fix template to say what a readiness failure actually does (held out of
Service endpoints, *not* restarted), with a pointer to yours. The docs said
`ReadinessProbeFailing (also covers liveness)`; that annotation is gone and both are listed
separately.

So your PR did not just add a playbook, it split a conflation that had been sitting in the library.
That is the more valuable half of it.

**Guards added:** `TestProbeDetectorsDoNotCrossFire` asserts all four directions — liveness fires
on liveness, liveness does *not* fire on readiness, readiness fires on readiness, readiness does
*not* fire on liveness. The two negative cases fail against the original files.

One note for next time, and it is a repo shortcoming rather than yours: `triggers:` and `detect:`
are two independent features, and nothing in the suite asserted a compiled `detect:` predicate can
fire — which is how both this and #127 went green. I am adding that requirement to the playbook
issue template.

---

Stepping back from this PR specifically — this batch alone is **twelve** merged PRs (#117, #118,
#119–#126, #127, #128), which takes you to **seventeen** in total after #57, #65, #70, #110 and
#112. That is by some distance the largest body of outside work KubeIntellect has, and it spans every layer of it:
file modes, lint families, playbooks, packaging, and now the entire `kq` cookbook. The examples
page went from documenting 2 of 10 subcommands to all 10 in a single sitting.

I have added `doc` to your entry in `.all-contributorsrc` and rewritten your line in the README's
contributors table to cover this batch.

One optional thing, and genuinely no expectation attached — it changes nothing here either way. If
you would like to post about any of this (LinkedIn, X, Mastodon, a Kubernetes community, a blog,
wherever you are active), please do. It is your work and the credit is yours, and "documented the
full CLI surface and shipped two failure-detection playbooks to an open-source AI SRE for
Kubernetes" is a concrete, specific thing to point at — much more useful to show than a commit
count. If you tag the repo I will amplify it from the project's side, and I am happy to review a
draft or confirm any technical detail before you post. If you would rather not, that is completely
fine and I will not bring it up again.

If you want something to pick up next, two that suit you: **#20** (the detector engine stores
PromQL in playbooks but never executes it — the `promql:` blocks you have now written three of are
currently documentation), and **#71**, which needs a judgement call more than code.

MSKazemi:
Following up on my note above: I said I would put the "prove the `detect:` predicate fires" requirement in the playbook issue template. There is no playbook issue template — #94/#95 were hand-written — so it landed in the two places an author will actually hit instead, in #130:

- **`.github/PULL_REQUEST_TEMPLATE.md`** — a playbook-specific checklist item, so it is in front of you while you open the PR.
- **`v4/docs/agent-behaviors.md`**, immediately under the schema — a warning box spelling out that `kind:` is the observation channel (`Pod` / `Event` / `Node`) and never the Kubernetes object, that `involved_kind:` is how you narrow the subject, and that a shared reason like `Unhealthy` needs a `message_regex` co-condition.

Both point at the tests from your PR as the worked example of the right shape. Sorry for the imprecision in the original comment.

MSKazemi:
You have merged seventeen pull requests across my repositories — more than anyone else — and the worked examples in particular are what made the CLI documentation usable rather than merely present. Thank you, genuinely.

I have a newer project, [IDKMesh](https://github.com/MSKazemi/idkmesh), that is at roughly the stage kubeintellect was in before you arrived: real infrastructure on `main`, and documentation that nobody outside my own head has ever tested. It is research on whether humans and AI agents can verify each other's work well enough to be trusted.

I have just opened a batch of bounded newcomer issues, so if you want first pick they are here: [`good first issue`](https://github.com/MSKazemi/idkmesh/labels/good%20first%20issue). [#402](https://github.com/MSKazemi/idkmesh/issues/402) and [#397](https://github.com/MSKazemi/idkmesh/issues/397) are closest to the kind of work you did here.

No obligation — you have already contributed plenty.

hariomlohardev:
I am right know some what bussy in study's but i will do as much as contribution i can do in opensoure i love opensoure and you may checkout my projects also [hariomlohardev/peek](https://github.com/hariomlohardev/peek/) 
😊

MSKazemi:
@hariomlohardev — studies first, genuinely. That is not politeness; it is the correct priority order, and open source will still be here afterwards. Please do not read any of my earlier message as a claim on your time.

I had a proper look at [**peek**](https://github.com/hariomlohardev/peek), and a few things stood out.

**"htop for codebases"** is a good pitch because it tells me the *interaction model*, not just the feature — I immediately know it is a live thing I sit in rather than a report I read once. And `pip install peek-code && peek .` on the second line is the right call: the fastest way to lose a reader is to make them scroll to find out how to try it.

The detail I actually admired is a small one. Your demo is generated **by code** — `python -m peek.tools.gen_demo`, Pillow, no `vhs` — with SVG and HTML alongside the GIF. Most projects record a terminal demo by hand, and then it silently rots the moment the UI changes, because re-recording is annoying enough that nobody does it. Yours regenerates. I have been bitten by exactly that rot in this repo (a README demo that had drifted from the real output), so I recognise the problem you designed around. That is the same instinct that made your work here good: you kept fixing the *cause* rather than the symptom — the `.gitignore` glob in [#117](https://github.com/MSKazemi/kubeintellect/pull/117) rather than the missing file the issue blamed, and the executable bit in [#70](https://github.com/MSKazemi/kubeintellect/pull/70) rather than silencing the rule.

One friendly suggestion, from having made the mistake myself: your README carries a `tests-262 passed` badge as a **static image**. Ours drifted 5× out of date before I gated it, and a wrong count is worse than no count because people use it to judge whether their own run was complete. If it is ever cheap to point that at a real CI badge, it will save you a small embarrassment later.

## On contributing here

There is no backlog waiting on you and no expectation attached. When you do have a free evening and *want* one, these are the live ones:

- **[#13](https://github.com/MSKazemi/kubeintellect/issues/13)** — another failure playbook. You have written three; you would be fastest at this.
- **[#113](https://github.com/MSKazemi/kubeintellect/issues/113)** — `brew audit --strict` plus a real `brew install` against the kube-q formula. You did the Homebrew style pass in [#118](https://github.com/MSKazemi/kubeintellect/pull/118), so you already have the context loaded.
- **[#173](https://github.com/MSKazemi/kubeintellect/issues/173)** — a real bug, and more interesting than it sounds.

And if the answer is "not this month", that is a completely fine answer that needs no reply.

---

Two small reciprocal things, both optional. If you ever write up **peek** — how the code-generated demo works would make a genuinely good short post, because the rot problem is universal and most people have not thought of that fix — send it my way and I will share it. And if you think KubeIntellect is worth other people's attention, a ⭐ or a mention wherever you are helps a one-maintainer project more than it looks; you have already given it seventeen merged PRs, so please treat that as an invitation and never an ask.

Good luck with the exams. 😊

PR #133, opened 2026-08-14
fix(cortex): retry triage when the reply is not valid JSON

### LINKED ISSUE #22: Add a bounded retry to triage JSON parsing

uuzzrm:
## What & why

Closes #22

The triage tier is told to answer in strict JSON, but when the model actually returns something unparseable, we just swallowed it and defaulted to investigate — no second chance. For a tier whose whole job is classifying the request, that silently throws away the user's intent on a single bad reply.

This adds a bounded repair loop: if the reply doesn't parse, we feed it back with a corrective hint and retry, up to 3 total calls. Only after that do we fall back to investigate, same as before. The old behavior (exception -> immediate fallback) is preserved.

To make the loop work I split the parser in two: `_parse_triage_json_strict` returns `None` when the reply isn't usable JSON (so the loop can tell "bad reply" from a deliberate `investigate` answer), and `_parse_triage_json` keeps its old lenient contract for existing callers.

## Type of change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📝 Documentation
- [ ] 🧹 Refactor / chore
- [ ] ⚡ Performance
- [ ] 🔒 Security

## Scope

## Checklist

- [x] Docs updated if behavior/CLI/flags changed

## How to test

```bash
cd v4
uv run python -m pytest tests/test_cortex.py -q   # 20 passed
```

The new `TestTriageRetry` class covers: malformed-then-valid reply (retries once and uses the valid plan), repair hint actually reaching the retried prompt, and exhausting all attempts falling back to investigate. The malformed-then-valid test fails on the pre-fix code (single call, wrong mode) and passes after.

Full gates run locally: `ruff check` clean, `mypy` 0 errors across 171 files, `make check-syntax` OK. Note the full `pytest tests/` run has 46 failures (playbooks/detectors/doc-claims), but they fail identically on a clean checkout of `main` — I diffed the two failure sets and they're the same list, so nothing here adds a new failure.

## Notes for reviewers

The retry appends the model's malformed reply plus the hint as messages, so the second call sees what went wrong rather than guessing. 3 attempts felt like the right bound: enough for a real recovery, small enough that a broken model doesn't multiply latency much (the tier is the cheap small model anyway).

AI assistance: I used AI tooling to help draft the repair loop and the tests, and reviewed the diff and gate output before submitting.

### DIFF
--- a/v4/packages/kubeintellect-server/app/cortex/graph.py
+++ b/v4/packages/kubeintellect-server/app/cortex/graph.py
@@ -104,12 +104,20 @@ diagnosed), what was done or recommended, and concrete next steps. Be precise
 and cite the actual object names/namespaces/exit codes you observed. No tool
 calls — answer only."""
 
+# Triage repair loop (#22): the triage tier answers in strict JSON; when the
+# reply does not parse, we feed it back with a corrective hint and retry before
+# falling back to the investigate default.
+_TRIAGE_MAX_PARSE_ATTEMPTS = 3
+_TRIAGE_REPAIR_HINT = (
+    "Your previous response was not valid triage JSON. Reply with ONLY a JSON "
+    'object matching the schema: {"mode": "chat" | "investigate", '
+    '"plan": ["step 1", ...]}. No prose, no markdown fences.'
+)
+
 
 # ── Nodes ─────────────────────────────────────────────────────────────────────
 
 async def triage(state: CortexState, config: RunnableConfig) -> dict:
-    from app.cortex.models import get_triage_llm
-
     session_id = state["session_id"]
     await emit(session_id, StatusEvent(
         phase="analyzing", message="Triaging request…", session_id=session_id,
@@ -128,12 +136,7 @@ async def triage(state: CortexState, config: RunnableConfig) -> dict:
         SystemMessage(content=_TRIAGE_SYSTEM + "\n\n" + "\n\n".join(p for p in context_parts if p)),
         HumanMessage(content=user_text),
     ]
-    try:
-        reply = await get_triage_llm().ainvoke(prompt, config)
-        parsed = _parse_triage_json(str(reply.content))
-    except Exception as exc:
-        logger.warning(f"cortex.triage failed ({exc}) — defaulting to investigate")
-        parsed = {"mode": "investigate", "plan": []}
+    parsed = await _triage_with_repair(prompt, config)
 
     mode = parsed.get("mode", "investigate")
     steps = [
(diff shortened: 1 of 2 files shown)


MSKazemi:
Merged — thank you. This one is a better fix than the issue asked for.

#22 asked for a bounded retry. The thing that makes it work is a design decision you made that the issue never mentioned: splitting the parser so that **"unusable reply" and "investigate" stop being the same value**. The old `_parse_triage_json` returned `{mode: investigate}` for both, which is precisely why a retry was impossible to write — there was nothing left to branch on. Returning `None` is what created the seam.

Where it lives now: [`_triage_with_repair`](https://github.com/MSKazemi/kubeintellect/blob/a47c83d46e268275f8adef830f93eb138518778e/v4/packages/kubeintellect-server/app/cortex/graph.py#L165-L200).

**What I verified before merging:** red-green confirmed — collapsing `_TRIAGE_MAX_PARSE_ATTEMPTS` from 3 to 1 fails exactly `test_retries_once_when_first_reply_is_malformed` and `test_repair_hint_passed_on_retry`, and leaves the other two green. ruff clean, mypy 0/171, 1024 server tests. All 9 required checks green on `891043a`.

**What I changed on top** (#153 and #152 — full disclosure, none of it a criticism of the patch):

1. Removed `_parse_triage_json`. Once the repair loop landed it had no production caller, and its docstring said it was kept "for callers that want the old lenient behaviour" — there were none. The four existing parsing tests now assert against the strict parser directly.
2. Made the strict parser reject a `plan` that is not a list. `triage()` does `(parsed.get("plan") or [])[:6]`, so a **string** plan sliced into six single characters and emitted six one-character PlanSteps. That is pre-existing — the old parser was equally lax — but your PR built the right place to fix it, and rejecting it now sends the reply back through *your* loop, which is what the loop is for.
3. Capped the echoed malformed reply at 2 000 chars, so a pathological reply cannot be re-sent on every remaining attempt.

**And a heads-up, because it was my mistake and not yours:** merging this together with #138 turned `main` red. Both PRs were branched 31 commits back and both added tests, so neither CI run ever saw the other's — the doc-claims gate compares a hard-coded test count in `AGENTS.md` against a live collect, and 1040 became 1047. Nothing wrong with either patch; our gate simply cannot see a number that only two merges *together* move. Fixed in #152, and I have noted the gate itself as a follow-up.

Credited in CHANGELOG.md and in the README contributors table, and `test` added to your all-contributors entry — both PRs arrived with real regression tests, which is the part that made them mergeable at this speed.

— Mohsen

MSKazemi:
The bounded repair loop was the right call — feeding the malformed reply back with a corrective hint and retrying up to a limit, rather than either giving up on the first parse failure or retrying forever. The test that fails on the pre-fix code and passes after is what made it reviewable. Same for reading the playbook YAML as UTF-8 regardless of locale; that class of bug is invisible until it is on someone else's machine.

You have now fixed the agent boundary in this repo five times, which is why I am mentioning [IDKMesh](https://github.com/MSKazemi/idkmesh). It is a research project on whether a swarm of humans and agents can verify each other's work well enough to be trusted — and the failure mode you keep fixing here, an agent's output being *plausible* rather than *correct*, is precisely what it is trying to measure rather than assume away.

Honest about the stage: 1 star, no contributors but me. There is real infrastructure on `main` — task and verification contracts, A2A/MCP bindings, a simulation suite — and the README is explicit about what is not yet finished.

Newcomer issues are labelled [`good first issue`](https://github.com/MSKazemi/idkmesh/labels/good%20first%20issue) if you ever want a look. Thanks for the fixes here regardless.

PR #134, opened 2026-08-14
playbook: add DNS resolution failure (#96)

### LINKED ISSUE #96: playbook: in-cluster DNS resolution failing

uuzzrm:
## What & why

Closes #96

DNS is the most common reason a service is up but the app can't reach it, and it didn't have a playbook. This adds one for in-cluster resolution failures, covering both directions the issue calls out: CoreDNS itself unhealthy, and a workload with a broken `dnsPolicy`/`dnsConfig` or a NetworkPolicy blocking egress to port 53.

## Design notes

The failure surfaces in **application logs** ("no such host", "Temporary failure in name resolution", `dial tcp: lookup ... on 10.96.0.10:53`) and in CoreDNS `SERVFAIL` lines. Neither is visible to the kubectl `--watch` sensorium — pod STATUS stays Running and no Warning event is emitted — so there is nothing to compile into a watch predicate. Same call `NetworkPolicyBlocking` makes for the same reason: detection stays prompt-side (`detect: null`), with event-message triggers for the log signatures that do reach the snapshot.

The investigation steps are ordered the way I'd actually debug it: confirm CoreDNS is up → read its logs for SERVFAIL → verify the kube-dns Service/endpoints → check the workload's `dnsPolicy`/`dnsConfig` → rule out a policy dropping port 53 → probe from inside the cluster to separate cluster-wide from workload-local.

## Tests

Added trigger tests in `v4/tests/test_playbooks.py` (match on the canonical lookup error and on "Temporary failure in name resolution"; explicitly does *not* cross-fire on plain i/o timeouts, which NetworkPolicyBlocking owns). The playbook count claim is bumped to 24 everywhere the doc-claims gate checks.

Gates run locally:

- `uv run python scripts/check_doc_claims.py` — passes
- `uv run python -m pytest tests/ -q` — 1023 passed
- `uv run ruff check packages/kubeintellect-server/app/ packages/ki-protocol/` — clean
- `uv run mypy packages/kubeintellect-server/app packages/ki-protocol packages/kube-q/kube_q` — 0 errors
- `scripts/check-file-modes.sh`, `scripts/check-syntax-warnings.py` — OK

I don't have a live cluster to reproduce the failure against, so the `expected_evidence` lines that read like diagnosis guidance (upstream resolver in the Corefile, node resolv.conf) are reasoned from the schema and the existing playbooks rather than observed output.

AI assistance: I used AI tooling to help draft the YAML and the trigger tests, and reviewed the diff and gate output before submitting.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -29,7 +29,7 @@ The `v4/` platform is the current line and the only one you should deploy.
 - Conversational diagnosis correlating **kubectl + Prometheus + Loki**.
 - **Human-in-the-loop approval gate** with RBAC (superadmin / admin / operator / readonly) on
   every mutating action, enforced server-side at the mutating chokepoint.
-- **23 declarative failure playbooks** (`detect → investigate → remediate`), 20 of which compile
+- **24 declarative failure playbooks** (`detect → investigate → remediate`), 20 of which compile
   to zero-token detectors.
 - **Zero-token detection** — a detector engine over a kubectl `--watch` sensorium fires
   findings without invoking the LLM.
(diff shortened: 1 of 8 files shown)


MSKazemi:
Reopening the loop on this one properly: **your commit is now in `main`** (4a41c71, via #187), carried unmodified with your authorship.

This PR was closed on 2026-08-15 in a batch with #135 and #137 **without ever being merged and without a comment explaining it**. Those two were adopted; this one was not, and nobody told you. Full write-up on #96. That was my error, not a problem with the contribution — the playbook, the three tests, and your call to leave `detect:` null were all correct.

Thank you for the work, and for your patience with the delay.

Issue #75, opened 2026-08-10
Clear the ruff 0.16 backlog (439 findings), then lift the pin — replaces the prematurely-closed #64


MSKazemi:
`ruff` is pinned `<0.16` in `v4/pyproject.toml:25`, and `ruff>=0.16` is in `.github/dependabot.yml`'s ignore list. That pin is deliberate — 0.16 turned on rules that report a large backlog — but it means **CI is blind to every rule 0.16 added**, and the pin cannot be lifted until the backlog is cleared.

#64 tracked this work and was **closed prematurely** when [#70](https://github.com/MSKazemi/kubeintellect/pull/70) cleared one of the four rule families. The rest was never re-filed, so it has been untracked since. This issue replaces it.

## Current state, measured today (2026-08-10)

```
$ uvx ruff@latest check v4 --statistics
```

`ruff 0.16.2`, **439 findings across `v4/`**:

| Count | Rule | Autofix | Notes |
|---:|---|:--:|---|
| 142 | `BLE001` blind-except | — | Judgment call per site; the biggest chunk |
| **95** | **`UP045` non-pep604-annotation-optional** | `[*]` | ⚠️ **SAFETY TRAP — read below before touching** |
| 75 | `I001` unsorted-imports | `[*]` | Mechanical — split out as its own issue |
| 29 | `PLW1510` subprocess-run-without-check | — | Each needs a real decision about failure handling |
| 17 | `UP035` deprecated-import | `[-]` | |
| 11 | `RUF059` unused-unpacked-variable | — | |
| 8 | `SIM117` multiple-with-statements | — | |
| 7 | `S110` try-except-pass | — | Overlaps `BLE001` |
| ~55 | 25 further rules, ≤5 each | mixed | |

209 are autofixable. **That number is a trap, not an invitation** — see below.

## ⚠️ Never run `ruff check --fix` across this repo

**`UP045` rewrites `Optional[X]` → `X | None`. On an injected `RunnableConfig` parameter that silently disables RBAC and the human-in-the-loop approval gate.**

`langchain_core` decides whether to inject the run config by matching the annotation **by identity** (`type_ is RunnableConfig`). Widen it and the parameter stops being injected: the tool receives `config=None` and loses `user_role` (RBAC) and `hitl_bypass` (the HITL gate). **Every test still passes and the LLM-visible tool schema is unchanged**, so it is invisible exactly where you would look first.

Affected files:

[...]

PR #135, opened 2026-08-14
chore(server): modernize typing imports and annotations (UP006, UP035)

uuzzrm:
## What & why

This is the typing-modernization slice of the ruff 0.16 backlog from #75 (item 3, `UP006` / `UP035`). The server package still pulls `List`/`Type`/`Deque` out of `typing` and uses `typing.{Callable, Awaitable, Iterable, AsyncIterator}` where the stdlib equivalents are the modern spelling.

All 16 files already import `from __future__ import annotations`, and every name being moved is only used in annotations or type aliases, so this is purely mechanical with no runtime behavior change:

- `typing.List` / `typing.Type` -> `list` / `type` (`core/llm.py`)
- `typing.Deque` -> `collections.deque` (`memory/change_ledger.py`, `memory/fleet_exchange.py`)
- `typing.Callable` / `Awaitable` / `Iterable` / `AsyncIterator` -> `collections.abc` (13 files)

Refs #75 (this family; the rest of the backlog stays for its own PRs).

## Type of change

- [x] Refactor / chore

## Scope

- [x] This change is scoped to a version whose contributions are open (`v4/`)

## Checklist

- [ ] New behavior has both a happy-path and an error-path test - no behavior change, annotation-only
- [x] `uv run pytest` passes locally - see Notes
- [ ] Docs updated if behavior/CLI/flags changed - none

## Notes for reviewers

Gates run from `v4/`:

| Gate | Result |
|---|---|
| `uv run ruff check packages/kubeintellect-server/app/ packages/ki-protocol/` | clean |
| `uv run mypy packages/kubeintellect-server/app packages/ki-protocol packages/kube-q/kube_q` | 0 errors (171 files) |
| `uvx ruff@latest check v4 --select UP006,UP035` | clean - family cleared |
| `uv run python -m pytest tests/ -q` | 1020 passed with `PYTHONUTF8=1` |

Two environment notes, both pre-existing and unrelated to this change:

1. **Windows locale vs the playbook YAMLs.** On this machine the server suite fails a batch of playbook tests unless `PYTHONUTF8=1` is set, because the playbook loader reads the YAML files without an explicit encoding and this Windows locale defaults to GBK (the files are UTF-8). It reproduces on a clean checkout of `main` with zero changes, and CI runs on Ubuntu so it never sees it. I'll send a small separate PR for the loader (`path.read_text(encoding="utf-8")`).
2. **kube-q file-mode test.** 311/312 - the one failure is `test_load_or_create_user_id_explicit`, which asserts a POSIX mode (`0o600`) that Windows does not enforce. Passes on Linux CI.

Used an AI coding assistant for parts of the mechanical edit; the change itself is small and I ran all the gates above locally. Happy to explain any line.

### DIFF
--- a/v4/packages/kubeintellect-server/app/agent/playbooks/loader.py
+++ b/v4/packages/kubeintellect-server/app/agent/playbooks/loader.py
@@ -18,9 +18,9 @@ Schema (one .yaml file per playbook):
 from __future__ import annotations
 
 import re
+from collections.abc import Iterable
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Iterable
 
 import yaml
 
--- a/v4/packages/kubeintellect-server/app/api/v1/endpoints/chat_completions.py
+++ b/v4/packages/kubeintellect-server/app/api/v1/endpoints/chat_completions.py
@@ -20,7 +20,7 @@ import asyncio
 import json
 import time
 import uuid
-from typing import AsyncIterator
+from collections.abc import AsyncIterator
 
 from fastapi import APIRouter, HTTPException, Request
 from fastapi.responses import StreamingResponse
--- a/v4/packages/kubeintellect-server/app/autonomy/promotion_engine.py
+++ b/v4/packages/kubeintellect-server/app/autonomy/promotion_engine.py
@@ -14,8 +14,9 @@ against the P0 replay fixtures.
 
 from __future__ import annotations
 
+from collections.abc import Callable
 from dataclasses import dataclass, field
-from typing import Callable, Optional
+from typing import Optional
 
 from app.autonomy.promotion_stats import (
     Event,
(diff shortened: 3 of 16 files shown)


MSKazemi:
@uuzzrm — flagging something I should have caught at the time, not asking you to explain anything.

This PR was green on every check and cleared exactly what it said it would: `UP006`/`UP035` across the 16 files you listed, verified clean with `uvx ruff@latest check v4 --select UP006,UP035`, mypy and both suites passing. You closed it about 40 minutes after posting about it on #75 — I don't know why, and I'm not assuming a reason; there's no review or comment from me on it either way.

Six days later I merged [#159](https://github.com/MSKazemi/kubeintellect/pull/159), which cleared the same `UP006`/`UP035` family as part of a larger sweep (`UP045`, `UP037`, `I001`, and several others), with no reference to this PR or to you. I want to be direct about what that means: the rule family you'd already cleared, correctly, got redone under my name without credit — not because your work was wrong, but because I wasn't tracking it against #75 when I picked that up.

If any of the specific diffs in #159 differ from what you had here in a way that matters to you, I'm glad to look at it. Either way, this was on me to connect, and I wanted you to hear that plainly rather than have it sit unremarked.

PR #137, opened 2026-08-14
fix(playbooks): read playbook YAML as UTF-8 so non-UTF-8 locales don't drop every playbook

### LINKED ISSUE #136: playbook loader fails on non-UTF-8 locales: all playbooks silently fail to load (Windows)
### Describe the bug

`app/agent/playbooks/loader.py` reads every playbook YAML with `path.read_text()`, which uses the platform default encoding. The YAML files are UTF-8 (they contain em-dashes and other non-ASCII in the investigation steps and fix templates), so on any locale whose default encoding is not UTF-8 — Windows with a Chinese (GBK/cp936) or Western (cp1252) locale, or a POSIX C locale — `yaml.safe_load(path.read_text())` raises `UnicodeDecodeError`.

The per-file `except` in `load_playbooks` swallows the error and logs `playbook_load_failed`, so the failure is silent from the user's perspective: the system keeps running with only the playbooks whose files happen to be pure ASCII (on this machine: `QuotaExceeded` and `WebhookAdmissionRejected` out of 23).

### To Reproduce

On Windows with a non-UTF-8 locale (or with `LC_ALL=C`):

```python
from app.agent.playbooks.loader import list_playbooks
print(list(list_playbooks()))
```

Observed: 2 playbooks load, and the log is full of

```
playbook_load_failed file=crashloopbackoff.yaml error=UnicodeDecodeError('gbk', ..., 'illegal multibyte sequence')
```

`v4/tests/test_playbooks.py` fails ~26 tests as a result (the matching tests assert playbooks like `NetworkPolicyBlocking`, `DeploymentRolloutStuck`, `HPANotScaling` are loaded). Setting `PYTHONUTF8=1` makes everything pass, which confirms the encoding is the only issue.

### Expected behavior

All playbook YAML loads on every platform. Reading an explicitly UTF-8 file should not depend on the locale.

### Suggested fix

```python
data = yaml.safe_load(path.read_text(encoding="utf-8"))
```

### Additional context

Not hit in CI because CI runs on Ubuntu where the default encoding is UTF-8. The files are authored as UTF-8 (the repo already ships an em-dash in the playbook docs), so pinning the read encoding is the correct fix rather than re-encoding the files.

uuzzrm:
## What & why

Fixes the playbook loader on non-UTF-8 locales. `loader.py` reads the playbook YAML with `path.read_text()` (platform-default encoding), but the files are UTF-8 — they contain em-dashes in the investigation steps and fix templates. On Windows (GBK/cp1252) or a POSIX C locale, every read raises `UnicodeDecodeError`, the per-file handler swallows it, and the agent silently runs with 2 of the 23 playbooks (only the ASCII-only ones load).

Closes #136

### Before

```
playbook_load_failed file=crashloopbackoff.yaml error=UnicodeDecodeError('gbk', ..., 'illegal multibyte sequence')
...
playbooks loaded: ['QuotaExceeded', 'WebhookAdmissionRejected']
```

### After

```
playbooks loaded: ['CrashLoopBackOff', 'ImagePullBackOff', 'NetworkPolicyBlocking', 'DeploymentRolloutStuck', ...]
```

## Type of change

- [x] Bug fix

## Scope

- [x] This change is scoped to a version whose contributions are open (`v4/`)

## Checklist

- [x] New behavior has both a happy-path and an error-path test - the existing `tests/test_playbooks.py` suite covers this (26 tests fail before, all pass after; `test_all_playbooks_load` asserts the full expected set)
- [x] `uv run pytest` passes locally - 1021 passed
- [x] `uv run mypy src` passes locally - 0 errors (171 files)
- [ ] Docs updated if behavior/CLI/flags changed - none

## Notes for reviewers

Verified on this Windows machine (GBK locale): the full server suite was failing ~46 tests before this change and passes 1021/1021 after, without any `PYTHONUTF8` or env tweaks. On Ubuntu CI this is a no-op (the default encoding is already UTF-8), which is why it has not surfaced there.

Small fix, and I ran all four gates above locally. Happy to explain.

Disclosure: developed with AI-assisted tooling (used as a coding aid); the change and tests were reviewed and all four gates were run locally.

### DIFF
--- a/v4/packages/kubeintellect-server/app/agent/playbooks/loader.py
+++ b/v4/packages/kubeintellect-server/app/agent/playbooks/loader.py
@@ -63,7 +63,7 @@ def _compile_trigger(raw: dict) -> Trigger:
 
 
 def _load_one(path: Path) -> Playbook:
-    data = yaml.safe_load(path.read_text())
+    data = yaml.safe_load(path.read_text(encoding="utf-8"))
     if not isinstance(data, dict):
         raise ValueError(f"Playbook {path.name} must be a YAML mapping")
     name = data.get("name")
--- a/v4/tests/test_playbooks.py
+++ b/v4/tests/test_playbooks.py
@@ -342,3 +342,34 @@ default     12s         Warning   FailedComputeMetricsReplicas   horizontalpodau
 def test_match_hpa_not_scaling_by_missing_cpu_request() -> None:
     matched = match_playbooks(HEALTHY_PODS, HPA_MISSING_CPU_REQUEST_EVENTS)
     assert "HPANotScaling" in matched
+
+
+def test_playbook_yaml_reads_as_utf8_regardless_of_locale(monkeypatch, tmp_path) -> None:
+    """Playbook YAML must decode as UTF-8 even when the platform default encoding is
+    not UTF-8 (Windows GBK/cp936, POSIX C locale). Regression test for #136."""
+    import pathlib
+
+    from app.agent.playbooks.loader import _load_one
+
+    real_read_text = pathlib.Path.read_text
+    encodings: list[object] = []
+
+    def recording_read_text(self, *args, **kwargs):
+        encodings.append(kwargs.get("encoding"))
+        return real_read_text(self, *args, **kwargs)
+
+    monkeypatch.setattr(pathlib.Path, "read_text", recording_read_text)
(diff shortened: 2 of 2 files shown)


uuzzrm:
Superseded by #138, which covers v2/v3/v4 as well. Closing this one.

uuzzrm:
Superseded by #138, which covers v2/v3/v4 as well. Closing this one.

PR #138, opened 2026-08-14
fix(server): read playbook YAML as UTF-8 regardless of locale

### LINKED ISSUE #136: playbook loader fails on non-UTF-8 locales: all playbooks silently fail to load (Windows)

uuzzrm:
## What & why

Closes #136

`path.read_text()` without an explicit encoding uses the platform default text encoding, which is not UTF-8 on Windows CP936/CP1252 locales or a POSIX C locale. The playbook YAML files are authored as UTF-8 (they contain em-dashes and other non-ASCII in the investigation steps and fix templates), so on such locales every non-ASCII playbook raised `UnicodeDecodeError`, and the per-file handler in `_load_all()` swallowed it — the server kept running with only the two pure-ASCII playbooks out of 23.

This pins the read to `encoding="utf-8"` in the loader in all three generations (`v2/`, `v3/`, `v4/` — the same defect is present in each) and adds a regression test that fails if a playbook read ever drops the explicit encoding.

Reproduced locally on a CP936 (Windows zh-CN) box: before the change only `QuotaExceeded` and `WebhookAdmissionRejected` load out of 23; after it, all 23 load.

## Type of change

- [x] Bug fix

## Scope

- Version directory touched: v2/, v3/, v4/

## Checklist

## Notes for reviewers

- The fix is one line per loader; the v2/v3 loaders carry the same defect as v4, so they get the same one-line change (their suites are not part of CI but pass locally).
- AI assistance was used for parts of the implementation and tests; the change was reviewed and verified locally before pushing.

### DIFF
--- a/v2/app/agent/playbooks/loader.py
+++ b/v2/app/agent/playbooks/loader.py
@@ -59,7 +59,7 @@ def _compile_trigger(raw: dict) -> Trigger:
 
 
 def _load_one(path: Path) -> Playbook:
-    data = yaml.safe_load(path.read_text())
+    data = yaml.safe_load(path.read_text(encoding="utf-8"))
     if not isinstance(data, dict):
         raise ValueError(f"Playbook {path.name} must be a YAML mapping")
     name = data.get("name")
--- a/v2/tests/test_playbooks.py
+++ b/v2/tests/test_playbooks.py
@@ -252,3 +252,30 @@ def test_match_node_not_ready_by_event_message() -> None:
     )
     matched = match_playbooks(HEALTHY_PODS, events)
     assert "NodeNotReady" in matched
+
+def test_playbook_yaml_is_read_as_utf8_regardless_of_locale(monkeypatch) -> None:
+    """Playbook YAML is always read as UTF-8, never the platform default.
+
+    Regression test for #136: on locales whose default text encoding is not
+    UTF-8 (Windows CP936/CP1252, POSIX C locale), a plain read_text() raises
+    UnicodeDecodeError on the em-dashes the playbooks contain, and the
+    per-file handler in _load_all() silently drops the playbook.
+    """
+    from pathlib import Path
+
+    from app.agent.playbooks import loader
+
+    real_read_text = Path.read_text
+    read_without_utf8: list[str] = []
+
+    def spy_read_text(self: Path, *args: object, **kwargs: object) -> str:
+        if kwargs.get("encoding") != "utf-8":
+            read_without_utf8.append(self.name)
+        return real_read_text(self, *args, **kwargs)
+
+    monkeypatch.setattr(Path, "read_text", spy_read_text)
+    loader._load_all()
+    assert read_without_utf8 == [], (
+        f"playbook YAML read without an explicit UTF-8 encoding: {read_without_utf8}"
+    )
+
--- a/v3/app/agent/playbooks/loader.py
+++ b/v3/app/agent/playbooks/loader.py
@@ -60,7 +60,7 @@ def _compile_trigger(raw: dict) -> Trigger:
 
 
 def _load_one(path: Path) -> Playbook:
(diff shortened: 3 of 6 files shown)


MSKazemi:
Merged — thank you. This is a good bug report *and* a good fix, and the two are not always the same person's work.

The part worth naming: you did not stop at "add `encoding`". You went and found that `_load_all()`'s per-file `except` was what turned a decode error into **silence** — the server coming up with zero playbooks and nothing in the logs that looked like a failure. That is the actual severity of this bug, and it is why it merged as-is.

The regression tests are the reason I could merge quickly. Asserting that `read_text` is *called with* `encoding="utf-8"` rather than trying to simulate a CP1252 locale is the right call — it tests the invariant instead of the environment, so it will still be true on a runner that has no exotic locale installed.

Where it lives now: [`loader.py#L66`](https://github.com/MSKazemi/kubeintellect/blob/a47c83d46e268275f8adef830f93eb138518778e/v4/packages/kubeintellect-server/app/agent/playbooks/loader.py#L64-L68).

**What I verified before merging:** red-green confirmed — reverting only the three `encoding="utf-8"` arguments fails exactly your new test and nothing else. ruff clean, mypy 0/171, 1021 server tests. All 9 required checks green on `63d8511`.

**One thing you could not have known, and it is our gap not yours:** CI runs only the v4 and kube-q suites, so the v2 and v3 regression tests you added are **not guarded by CI**. I ran them by hand for this merge — v2 25 passed, v3 12 passed. I have noted it as a follow-up; it is a real hole that your PR happened to expose.

Also worth flagging, and genuinely out of scope for #136: the same bug class is alive in roughly 25 other places — `cli.py`'s config reads and writes, `kube_q/cli/repl.py:99,121` handling `.env`, and `kube_q/core/session.py:90,95,100`. If you ever want a follow-up with a clear boundary, that sweep is one, and you already know the failure mode better than anyone.

Credited in CHANGELOG.md and in the README contributors table.

— Mohsen

PR #139, opened 2026-08-15
fix(context-fetcher): never report an all-clear from a truncated or failed snapshot

### LINKED ISSUE #140: Snapshot health flags can report a false all-clear when the pod listing is truncated or kubectl fails
### Summary

`context_fetcher` runs `kubectl get pods --all-namespaces` and `kubectl get events --all-namespaces` and hard-caps each listing at 8 000 chars. The health scan (`_scan_snapshot`) and the flags it feeds the coordinator (`snapshot_has_issues` / `snapshot_has_warnings`) are derived from that already-capped text, and the cap is silent:

1. **Truncation.** On a cluster with ~80+ pods the pod table overflows the cap and is cut mid-listing. A CrashLoopBackOff / ImagePullBackOff pod sitting past the cut is never seen — `snapshot_has_issues` comes back `False`, and the LLM is handed a partial pod table with no indication that it is partial.
2. **Fetch failure.** When kubectl itself fails (timeout, bad kubeconfig), `pods_out` is `"(unavailable: ...)"`, which parses as a header row with zero pods — `snapshot_has_issues` comes back `False` again.

For an incident-response tool this is a false all-clear: the cluster can be reported clean while an unhealthy pod is invisible. The same truncation path also lets `coordinator._verify_resolution` mark a fix as `resolved` from partial data.

### Expected

A truncated or unavailable listing should never produce an all-clear signal, and the LLM-facing snapshot should say the listing is incomplete.

### Proposed

Carry `truncated` / `unavailable` flags out of the fetch, treat them conservatively in `_scan_snapshot`, and append a visible note to the snapshot text. I have a working branch with tests and am opening a PR.

uuzzrm:
## What & why

Closes #140

The snapshot health flags (`snapshot_has_issues` / `snapshot_has_warnings`) were derived from kubectl output that had already been hard-capped at 8 000 chars — no marker, and no awareness that the cap had been hit. On a cluster with enough pods to overflow the cap, an unhealthy pod past the cut was invisible to both the scan and the LLM, so the coordinator could report the cluster as clean when it wasn't. A kubectl failure did the same for the pod listing (zero pods parsed → no issues).

- `_run_kubectl_snapshot` now returns a small `SnapshotOutput` (text + `truncated` / `unavailable` flags) instead of a bare string.
- `_scan_snapshot` treats truncated/unavailable listings conservatively — no all-clear from partial or failed data.
- The snapshot text the LLM sees carries an explicit note when a listing was cut short or never arrived (mirrors the existing marker in `run_kubectl`).
- `coordinator._verify_resolution` inherits the same conservatism, so a fix is no longer marked "resolved" from a partial listing.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Docs
- [ ] Refactor / chore
- [ ] Performance
- [ ] Tests

## Scope

- Version directory touched: v4/

## Checklist

- [x] New behavior has both a **happy-path** and an **error-path** test (`tests/test_context_fetcher.py`)
- [x] No mutating ops touched — the HITL/RBAC invariants are unaffected
- [x] Secret values are never logged or returned (n/a here)
- [x] `uv run pytest` passes locally — see gate output below
- [x] `uv run ruff check` passes locally
- [x] `uv run mypy` passes locally — 171 source files, 0 errors
- [x] Docs updated if behavior/CLI/flags changed — n/a, no user-facing behavior change

## Notes for reviewers

Gates run (all from `v4/`):

- `uv run ruff check packages/kubeintellect-server/app/ packages/ki-protocol/` — clean
- `uv run mypy packages/kubeintellect-server/app packages/ki-protocol packages/kube-q/kube_q` — 0 errors (171 source files)
- `uv run python -m pytest tests/ -q` — 1027 passed under `PYTHONUTF8=1`

One environment note: on my Windows box (GBK locale, code page 936) the default-locale run shows 46 failures in `tests/test_playbooks.py`. That is the pre-existing playbook loader encoding bug tracked in #136 — the loader uses `path.read_text()` without an explicit encoding, so only 2 of the playbooks load under a non-UTF-8 locale. It is unrelated to this change and green under a UTF-8 locale, which is what CI runs.

Heads-up: I used AI assistance to help draft and verify this change — happy to walk through any part of it in review.

### DIFF
--- a/v4/packages/kubeintellect-server/app/agent/nodes/context_fetcher.py
+++ b/v4/packages/kubeintellect-server/app/agent/nodes/context_fetcher.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import asyncio
+from dataclasses import dataclass
 import os
 import subprocess
 import time
@@ -15,6 +16,45 @@ logger = get_logger(__name__)
 
 _SNAPSHOT_MAX_CHARS = 8_000
 
+# Notes appended to the snapshot when a listing was cut short or never
+# arrived. Mirrors the truncation marker in tools/kubectl_tool.py: the LLM
+# must be told the list it is looking at is incomplete, never left to infer
+# an all-clear from partial data.
+_SNAPSHOT_TRUNCATED_NOTE = (
+    "[snapshot truncated: output exceeded 8 000 chars and was cut short — "
+    "this listing is incomplete and may hide unhealthy pods or warning "
+    "events. Narrow the scope with -n <namespace> or -l <label>.]"
+)
+_SNAPSHOT_UNAVAILABLE_NOTE = (
+    "[snapshot unavailable: kubectl could not fetch this listing — "
+    "treat the cluster state as unknown, not healthy.]"
+)
+
+
+@dataclass(frozen=True)
+class SnapshotOutput:
+    """One capped kubectl snapshot listing, plus flags that say whether an
+    all-clear derived from it can be trusted.
+
+    ``truncated``   — the output hit the 8 000-char cap, so rows beyond the
+        cut are unknown.
+    ``unavailable`` — kubectl could not be run at all (missing binary,
+        timeout, bad kubeconfig), so nothing was fetched.
+    """
+
+    text: str
+    truncated: bool = False
+    unavailable: bool = False
+
+    def with_note(self) -> str:
+        """The capped text with a visible note appended when it is incomplete."""
+        if self.unavailable:
+            return f"{self.text}\n\n{_SNAPSHOT_UNAVAILABLE_NOTE}"
+        if self.truncated:
+            return f"{self.text}\n\n{_SNAPSHOT_TRUNCATED_NOTE}"
+        return self.text
+
+
 # Pod phases that count as "healthy" — anything else flips snapshot_has_issues.
 # (Note: STATUS column from `kubectl get pods` mixes phases and reasons —
 #  e.g. "CrashLoopBackOff", "ImagePullBackOff". We treat any value not in this
@@ -22,14 +62,27 @@ _SNAPSHOT_MAX_CHARS = 8_000
 _HEALTHY_POD_STATUSES = frozenset({"Running", "Completed", "Succeeded"})
 
 
-def _scan_snapshot(pods_out: str, events_out: str) -> tuple[bool, bool, int]:
+def _scan_snapshot(
+    pods_out: str,
+    events_out: str,
+    *,
+    pods_truncated: bool = False,
+    pods_unavailable: bool = False,
+    events_truncated: bool = False,
+    events_unavailable: bool = False,
+) -> tuple[bool, bool, int]:
     """Return (has_issues, has_warnings, pod_count) by scanning kubectl output.
 
     Cheap line-based parse — no extra subprocess calls. The pod table format is:
         NAMESPACE   NAME   READY   STATUS   RESTARTS   AGE
     We index the STATUS column by header position to be robust to column widths.
+
+    The ``*_truncated`` / ``*_unavailable`` flags are conservative guards: a
+    listing that was cut short (or never fetched) must not be reported as
+    clean, because an unhealthy pod or warning event may sit beyond the cap.
+    Callers derive them from SnapshotOutput.
     """
-    has_issues = False
+    has_issues = pods_truncated or pods_unavailable
     pod_count = 0
 
     lines = pods_out.splitlines()
@@ -53,11 +106,19 @@ def _scan_snapshot(pods_out: str, events_out: str) -> tuple[bool, bool, int]:
         if status not in _HEALTHY_POD_STATUSES:
             has_issues = True
 
-    has_warnings = bool(events_out.strip()) and "No resources found" not in events_out
+    has_warnings = events_truncated or events_unavailable
+    if events_out.strip() and "No resources found" not in events_out:
+        has_warnings = True
     return has_issues, has_warnings, pod_count
 
 
-def _run_kubectl_snapshot(args: list[str]) -> str:
+def _run_kubectl_snapshot(args: list[str]) -> SnapshotOutput:
+    """Run a kubectl snapshot query, capping output at _SNAPSHOT_MAX_CHARS.
+
+    Returns a SnapshotOutput carrying the capped text plus ``truncated`` /
+    ``unavailable`` flags so callers never mistake a partial or failed
+    listing for an all-clear signal.
+    """
     kubeconfig = os.path.expanduser(settings.KUBECONFIG_PATH)
     env = {**os.environ, "KUBECONFIG": kubeconfig}
     try:
@@ -70,10 +131,13 @@ def _run_kubectl_snapshot(args: list[str]) -> str:
             shell=False,
         )
         out = proc.stdout or proc.stderr or ""
-        return out[:_SNAPSHOT_MAX_CHARS]
+        return SnapshotOutput(
+            text=out[:_SNAPSHOT_MAX_CHARS],
+            truncated=len(out) > _SNAPSHOT_MAX_CHARS,
+        )
     except Exception as exc:
         logger.warning(f"context_fetcher: kubectl {' '.join(args[:2])} failed: {exc}")
-        return f"(unavailable: {exc})"
+        return SnapshotOutput(text=f"(unavailable: {exc})", unavailable=True)
 
 
 async def context_fetcher(state: AgentState) -> dict:
@@ -85,7 +149,7 @@ async def context_fetcher(state: AgentState) -> dict:
         session_id=session_id,
     ))
 
-    pods_out, events_out = await asyncio.gather(
+    pods, events = await asyncio.gather(
         asyncio.to_thread(_run_kubectl_snapshot, ["get", "pods", "--all-namespaces"]),
         asyncio.to_thread(_run_kubectl_snapshot, [
             "get", "events", "--all-namespaces",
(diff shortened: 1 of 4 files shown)


MSKazemi:
Thank you for this — the bug is real, the diagnosis is right, and I am not merging it yet. Two reasons, and the second one is the useful one.

## 1. It does not yet catch the failure it is aimed at

\`_run_kubectl_snapshot\` never checks \`proc.returncode\`. \`subprocess.run\` does **not** raise on a non-zero exit, so \`unavailable\` only fires on the exception path — missing binary, timeout, bad kubeconfig path. The common failures do not go through there: \`Unauthorized\`, connection refused, an expired token. Those exit non-zero, \`proc.stdout or proc.stderr\` picks up the **stderr text**, and it gets parsed as a pod table.

I ran your branch against exactly that case — kubectl exiting 1 with `error: You must be logged in to the server (Unauthorized)`:

```
unavailable flag : False
truncated  flag  : False
has_issues       : False | pod_count: 0
=> ALL-CLEAR FROM A FAILED READ
```

Which is the thing your title promises to prevent. Your truncation half works correctly — that one I verified fires.

There is a second variant worth knowing about: a real connection failure prints as three lines, and the two `E0820 …` lines have enough whitespace-separated columns to be counted as pods, so you get `pod_count=2` invented out of an error message.

## 2. A maintainer rework of the same function is in flight

This is on me, not you — it was not visible from outside, and you had no way to know. I have an unfinished local rework of `_scan_snapshot` covering this same bug, and it takes the **opposite** design decision on the central question:

- **Yours:** `has_issues = pods_truncated or pods_unavailable` — a failed read reports *issues found*.
- **Mine:** `if not pods_ok: return False, False, 0` — a failed read reports *nothing known*, and callers distinguish "unknown" from "clean" via the flag they passed in.

Both are defensible. Yours fails loud, which is the safer default in isolation; mine keeps `has_issues` meaning strictly "I saw a problem" and moves the unknown-state signal to a separate channel, because `has_issues` also drives user-facing wording and I did not want an unreachable cluster described as an unhealthy one. **I have not decided yet**, and I am not going to pretend the decision is made just to close this out.

## What happens next

I am keeping this open, not closing it. The concrete blocker is one design call that is mine to make, and I will make it and come back here — not leave it drifting.

If you would like to keep it, the single change that would make this fix the real bug is checking `proc.returncode` in `_run_kubectl_snapshot` and setting `unavailable` from it, plus a test using a fake process with `returncode=1` and error text on stderr. Say the word and it is yours. If you would rather not spend more time on a PR whose seam I might rewrite underneath you, that is completely fair — say so and I will take it from here with your name on it either way. Your call, and neither answer is the wrong one.

One unrelated nit for whichever version lands: `context_fetcher.py` ends without a trailing newline in this branch.

Separately — #133 and #138 are both merged, with replies on each. And if you want #96 back, it is unclaimed and unassigned again.

— Mohsen

MSKazemi:
The design call is made. I said I would come back rather than leave this drifting, so here it
is, and it does not go your way — but the outcome for you does not change.

## The decision

**A failed read reports *nothing known*, not *issues found*.**

```python
if not pods_ok:
    return False, False, 0
```

`has_issues` keeps meaning strictly "I saw a problem", and the unknown-state signal moves to its
own channel that callers read separately.

The reason is the one I flagged when I laid the two options out: `has_issues` does not only
gate logic, it drives user-facing wording. Under your version an expired token or a network
blip gets described to the user as an unhealthy cluster. That trades one wrong answer for a
different wrong answer, and this project's whole position is that a confident wrong answer is
the failure mode it cannot ship. "I could not read the cluster" and "the cluster has problems"
have to stay distinguishable all the way to the sentence the user reads.

Your version fails loud, which is genuinely the safer default in isolation, and if `has_issues`
were purely internal I would have taken it. It is not, and that is the whole of the difference.

## What this means concretely

The rework is written against `main` and carries both halves — the `proc.returncode` check that
your branch is missing (`subprocess.run` does not raise on a non-zero exit, so `Unauthorized`
and connection-refused never reached your `unavailable` path), and the truncation handling that
your branch got right and I verified fires. It also stops an error message being parsed as a
pod table, which is how you could get `pod_count=2` invented out of two `E0820 …` stderr lines.

**Your name is on it either way.** That was true when I said it and the decision does not
change it — the truncation analysis, the `SnapshotOutput` shape, and the observation that
`_verify_resolution` inherits the same conservatism are all yours, and the fix carries them
whatever seam it sits on.

## Your call, and both answers are fine

1. **You take it.** Rework this branch onto the decided semantics — `if not pods_ok: return
   False, False, 0`, the `returncode` check, and a test with a fake process at `returncode=1`
   with error text on stderr. I will review it the day it lands.
2. **I take it**, and this PR closes as superseded with the credit intact and your name on the
   commit.

No deadline that costs you anything: say which, whenever. If I hear nothing in about a week I
will assume (2) rather than let this sit open for another month — that is not a penalty, it is
me not repeating the mistake of leaving it drifting.

## Two other things

The trailing-newline nit still applies to whichever version lands.

And separately: you are on the
[Contributors table](https://github.com/MSKazemi/kubeintellect#contributors) and in
`.all-contributorsrc` as `doc`, `security`, `code`, `test` — for #105, #111, #116, #133 and
#138. The roster was audited today and four people who had contributed were on no list at all
([#169](https://github.com/MSKazemi/kubeintellect/pull/169)); yours was already correct, and I
am mentioning it only because I checked.

MSKazemi:
@uuzzrm — an overdue update, and this time with a concrete way out rather than another "it is with the maintainer".

First, the honest part: this PR has been open since 15 August, blocked on a decision I said in [#140](https://github.com/MSKazemi/kubeintellect/issues/140#issuecomment-3210559873) was mine to make, and then did not make. Separately, I discovered today that your #134 was closed unmerged in a batch and never landed — [written up on #96](https://github.com/MSKazemi/kubeintellect/issues/96), now adopted in #186. Two threads of yours stalled for the same reason: I was not closing loops. I am sorry, and neither one reflects on the work.

**The finding here remains the most serious one anyone has reported against this project.** Safety invariant #1 is "never fabricate an answer", and a false all-clear from a truncated or failed read is precisely that failure — with the aggravating detail I added on #140, that a connection error's stderr has enough whitespace-separated columns to be *counted as pods*, so the tool can invent `pod_count=2` out of an error message.

## The decision, and a way to have both sides of it

To restate where we were. The disagreement was one line:

- **Yours:** `has_issues = pods_truncated or pods_unavailable` — a failed read reports *issues found*. Fails loud, which is the safer default in isolation.
- **Mine:** `if not pods_ok: return False, False, 0` — a failed read reports *nothing known*, because `has_issues` also drives user-facing wording and I did not want an unreachable cluster described as an unhealthy one.

Re-reading the consumer today, I think the disagreement was never real — we were both protecting different properties of the same flag, and it can carry only one. Here is what the flag actually feeds, in `coordinator.py`:

```
When the user asks a LIST-SHAPED, READ-ONLY question AND issues=false AND
warnings=false AND the snapshot is fresher than Ns:
  - Prefer answering directly from the snapshot.
```

So `has_issues=false` is not really the claim "the cluster is healthy". It is the licence to **skip the fresh read and answer from cached text**. That is the thing that must never be granted from partial data — and it is a *different* proposition from "unhealthy workloads were observed", which is what the wording needs.

And the branch that does this correctly already exists a few lines above, for the unavailable case:

```
**The cluster snapshot is UNAVAILABLE — the read failed.** It reports no pod
count and no health flags, because none are known. An unavailable snapshot is
not an empty or healthy cluster.
- ALWAYS use a tool to fetch what you need; never answer from the snapshot.
```

**So the proposal is: keep the flags meaning what they say, and gate the shortcut on a third signal.**

- `has_issues` / `has_warnings` keep your conservative derivation over whatever *was* read, and keep meaning "unhealthy workloads observed" — so an unreachable cluster is never *described* as unhealthy. (My constraint.)
- Add `snapshot_complete: bool` to `AgentState`, false when a listing was truncated **or** unavailable.
- The snapshot-sufficiency shortcut requires `issues=false AND warnings=false AND snapshot_complete`. A truncated listing routes into the existing UNAVAILABLE-style branch instead of the "answer from the snapshot" one. (Your constraint — no all-clear from partial data, and it now fails safe at the only place the flag can cause harm.)

That gets your safety property and my wording property with no trade, and it reuses a prompt branch that is already there and already correct.

**Which means the answer to your PR is yes** — your read of the bug, your `SnapshotOutput` carrying `truncated`/`unavailable` out of the fetch, your conservative `_scan_snapshot`, the explicit note in the LLM-facing text, and the `_verify_resolution` fix are all correct and all stay. The only change is that the truncation signal lands on its own field instead of being folded into `has_issues`.

## Practical next step — and no obligation

The branch has picked up conflicts since August (`context_fetcher.py` and `coordinator.py` have both moved). Two options, entirely your choice:

1. **You rebase and make that one change**, and I review it the day it lands. It stays your PR and your commit.
2. **I do the rebase and the `snapshot_complete` split on an integration branch**, carrying your commit unmodified with your authorship, exactly as I just did for #134 in #186 — and you review *me*.

I am genuinely happy with either, and option 2 exists specifically so that a three-week-old blockage that was my fault does not become extra work for you. Just say which. If I do not hear from you I will take option 2 in about a week, so that this cannot stall again — and you keep authorship either way.

One heads-up for whoever does it: `AGENTS.md`'s server-suite count is itself a gate now (`check_doc_claims.py`), so adding tests means updating that number or CI goes red on a check that does not run locally by default. It is currently in motion — two other PRs are in flight — so recollect it rather than copying a number.

Also, for the record on your environment note: you were right that the 46 `test_playbooks.py` failures were the #136 encoding bug and not your change. That bug is fixed in `main` now, by your own #138.

---

Thank you for your patience on this — three weeks of silence on a report this good is not what I want this project to be like, and the fix is on my side of the fence, not yours.

And an entirely optional aside, no strings attached to the review: this particular bug — *an incident-response tool reporting all-clear because its own output was truncated* — is a genuinely interesting write-up, and the class of failure generalises well beyond this repo. If you ever felt like posting about it, I would be glad to amplify it, and it would put your name on the finding rather than leaving it in a changelog entry. Likewise a ⭐ or a share if you think the project has earned it. Only if you want to. 🙏

MSKazemi:
Merged, and #140 is closed. Thank you — and I owe you an apology for how long this took.

Your commit is in `main`'s history, so GitHub has marked this PR merged and the authorship is yours where it counts: https://github.com/MSKazemi/kubeintellect/commit/079c96a77c4dd704c5cb8dae28af92cdda1ddd30

You were right, and it turned out to be worse than either of us wrote down. I reproduced it on `main` before touching anything:

```
full listing chars : 37398   (cap 8000)
ok flag            : True
CrashLoop in text? : False      <- the unhealthy pod is past the cut
any 'truncat' marker: False     <- nothing tells the model
has_issues         : False
pod_count reported : 85         <- real total is 401
```

A 401-pod cluster with a `CrashLoopBackOff` in it, reported to the operator as "85 pods, healthy". Both the number and the verdict invented. That was live in the released v2.5.0, and since every real cluster exceeds 8 000 characters of pod listing, it was firing in production rather than in theory.

**What changed since you filed it, and why your patch could not just be rebased.** `main` rewrote `_kubectl_snapshot` after you opened this, with the opposite return shape — `(ok, text)` plus `pods_ok`/`events_ok` short-circuiting — which closed the kubectl-failure half of your report in 2.4.0. That left your branch 188 commits behind with hard conflicts in the two files it touched. Asking you to rebase would have meant asking you to rewrite your commit against a seam that had moved twice underneath you, so I took it forward myself rather than sending it back.

**Where I departed from your design, and why.** You put truncation into `has_issues`. I kept your conservative derivation but gave it its own flag, `snapshot_complete`, because `has_issues` is also rendered to the user as "unhealthy workloads were observed" — and a cluster nobody could measure is not a cluster that is unhealthy. It gates the answer-from-the-snapshot shortcut instead, which is the place the harm actually happened. An incomplete snapshot now renders: *"NEVER answer 'how many pods', 'is the cluster healthy' … from this snapshot. Absence of a pod here is NOT evidence that it does not exist."*

**Three more defects came out of your report that neither of us had spotted.** Your finding generalised further than it looked:

1. The cap was deleting the blocked-namespace notice. `[Protected] N row(s) withheld — this listing is NOT the complete set` was appended at the end of the table and the cap ran afterwards, so on any cluster over the limit that sentence was sliced off — and the short listing read as a complete one, which is the exact thing the sentence exists to prevent. On every large cluster, namespace filtering was silent.
2. The cap cut mid-row, and the fragment still parsed: `default app-1 1/1 Runni` was counted as a pod whose STATUS is `Runni`, which is not a healthy phase. So a listing of nothing but `Running` pods reported an issue, out of a severed word.
3. The truncation marker I added was itself counted as a pod, because policy lines were matched by a `[Protected]` prefix and the marker begins `[truncated`. The notice saying pods were missing invented a pod, and an issue, out of itself.

Your eight tests are ported to the new API, and each of those three has one of its own.

Verified on the merged tree: ruff clean, mypy clean across 193 files, server suite 5576 passed, kq suite 749 passed, all four repo gates green, doc claims recollected to 5600. CI green on the merge commit.

You are already credited in `.all-contributorsrc` and in the changelog entry for the next release, which names you as the reporter.

The honest summary is that this sat for four weeks because I changed my mind twice about the design and did not come back to you quickly enough. The finding was good, it was real, and it was in shipped code the whole time. Thank you for filing it and for the tests — the negative cases are what made it safe to change the seam at all.

Entirely optional, and the merge and the credit stand regardless: if the project is useful to you, a star or a mention helps other people find it.

PR #109, opened 2026-08-10
test(kube-q): pin terminal rendering environment

floze-the-genius:
## What & why

Closes #106.

The kube-q suite inherited `COLUMNS`, `TERM`, and `NO_COLOR` from the contributor's shell. That changed Rich table wrapping, terminal-height behavior, and the theme used by the module-level renderer console, so unrelated tests could fail before a contributor changed any code.

This adds one suite-level pytest configuration that:

- pins `COLUMNS=120` and `TERM=xterm-256color` before test-module collection;
- clears `NO_COLOR` before collection so import-time consoles use the normal test theme;
- reapplies the same baseline with an autouse fixture before every test while still allowing individual tests to override it with `monkeypatch`;
- restores the invoking process environment when pytest exits.

No production code or assertion was changed.

## Type of change

- [x] 🐛 Bug fix
- [x] 🧪 Tests

## Scope

- Version directory touched: `v4/packages/kube-q/`
- [x] This change is scoped to a version whose contributions are open.

## Validation

On current `main` (`2c1f6765`), the clean one-variable-at-a-time RED matrix was:

- `COLUMNS=40`: 5 failed, 307 passed
- `COLUMNS=60`: 1 failed, 311 passed
- `COLUMNS=200`: 312 passed
- `TERM=dumb`: 1 failed, 311 passed
- `NO_COLOR=1`: 1 failed, 311 passed

With this commit, all five runs pass: **312/312 each** on Python 3.13.13. The full suite also passes **312/312 on Python 3.12.0** with all three hostile values applied simultaneously (`COLUMNS=40 TERM=dumb NO_COLOR=1`).

`make setup` passes all six locally runnable gates:

- Ruff: clean
- mypy: zero errors across 171 source files
- server suite: 995 passed
- kube-q suite: 312 passed
- file modes: clean
- syntax warnings: clean

## Checklist

- [x] The regression is proven by the unchanged assertions failing before and passing after the environment pin
- [x] No production write, HITL, RBAC, secret, CLI, or documentation behavior changes
- [x] All locally runnable CI gates pass
- [x] Commit is signed off (DCO)

## AI assistance

Substantial AI assistance was used to inspect the Rich/pytest import timing, implement the fixture, and prepare this PR. I reviewed the resulting code and ran every validation result reported above locally.

PR #161, opened 2026-08-22
Name an encoding on every text-mode read and write, and gate it

### LINKED ISSUE #156: read_text()/open() without encoding= in ~25 places — same bug class as #136
Follow-up to #136 / #138, which fixed the playbook loader.

That fix was correctly scoped to the bug reported. The **same bug class** is still live elsewhere: a bare `path.read_text()` or `open(path)` decodes using the *platform default* encoding, which is not UTF-8 on Windows (CP1252/CP936) or under the POSIX `C` locale. Any non-ASCII byte then raises `UnicodeDecodeError`.

Known sites:

**`v4/packages/kubeintellect-server/app/cli.py`** — config file reads and appends
`538`, `547`, `606`, `722`, `998`, `1131`, `1133`, `1187`, `1189`, `1301`, `1308`, `1486`, `1582`, `1649`, `1652`, `1684`, `1699`

**`v4/packages/kube-q/kube_q/cli/repl.py`** — `.env` read and write
`99`, `121`

**`v4/packages/kube-q/kube_q/core/session.py`** — user-id file
`90`, `95`, `100`

The `.env` and config paths are the ones that matter most: they are **read/modify/write** cycles on user files. A decode failure there is not a crash on read — it can mean writing back mangled content.

## What to do

Add an explicit `encoding="utf-8"` to every text-mode read and write above. This is mechanical and low-risk, but it is **not** a blind find-and-replace: skip binary-mode (`"rb"`/`"wb"`) calls, and check each write path actually round-trips.

Worth pairing with a guard so it does not come back — either a ruff rule if one covers this (`PLW1514` / `FURB129` family) or a small AST check in `scripts/` like the existing repo-specific checks.

Good first issue: the fix is obvious, the surface is enumerated above, and the test pattern already exists — see `test_playbook_yaml_is_read_as_utf8_regardless_of_locale` in `v4/tests/test_playbooks.py`, which spies on `Path.read_text` and asserts the encoding argument rather than trying to simulate a locale.

shaurya703:
Fixes #156.

## Scope: wider than the issue enumerates, deliberately

The issue lists ~22 sites in three v4 files. I did **62**, across the whole CI-linted scope — those three, plus `kube-q`'s tests and `check-event-parity.py`, `v4/tests`, `v4/scripts/fix_pr_probe.py` and the two demo scripts under `scripts/`.

The reason is the guard. A gate that exempts tests and probes is a gate with a carve-out, and "the places nobody looks" is where this class survives — it is how #136 reached a user in the first place. Frozen `v1/`-`v3/` are untouched, matching `check-syntax-warnings.py` and the scope statement in `ci.yml`.

Full count for the record: **119** sites repo-wide, 57 of them in the frozen generations and left alone.

## How the edits were made

By walking the AST and inserting the keyword after the last existing argument — not find-and-replace. Skipped: binary mode (`"rb"`/`mode="wb"`), calls already naming an encoding, and calls forwarding `**kwargs` where it may arrive at runtime. Every touched file was re-parsed before writing, and a second AST scan confirms zero remaining sites in scope.

The diff is mechanical and reads as one line per call:

```diff
-        with open(_USER_ID_FILE, "w") as f:
+        with open(_USER_ID_FILE, "w", encoding="utf-8") as f:
```

## The gate

`scripts/check-text-encoding.py`, same shape and scope as `check-syntax-warnings.py`: stdlib only, tracked `*.py` outside `v1`-`v3`, no virtualenv or network, independent of ruff and the `<0.16` pin.

On ruff: **PLW1514 does cover this rule**, and it would be the smaller change if it were reachable. It is not in the CI-linted scope today and enabling it repo-wide would also light up the frozen generations, so a scoped stdlib gate matches what the repo already does for #63. If the ruff upgrade lands and you would rather delete this script for a rule line, that trade seems right to me — the script says so in its header.

Wired into `make check-encoding` and the existing ci.yml job that runs the syntax gate.

`v4/tests/test_encoding_gate.py` mirrors `test_syntax_gate.py`, red cases included: the three call shapes are caught, binary mode and `**kwargs` are **not** flagged, the frozen filter is load-bearing, the scan is asserted non-vacuous, and `main()` returns 1 naming the file.

## Verification, and one caveat

- `./scripts/check-text-encoding.py` — 308 files, clean. Against a fixture with a bare `read_text()`, `write_text()` and `open("w")` it exits 1 and names all three.
- `./scripts/check-syntax-warnings.py` — still clean.
- `v4/tests/test_encoding_gate.py` — 6 passed.
- `kube-q` suite — **316 passed, 1 failed**, identical to the baseline I took before touching anything. The failure is `test_export_to_file`, an ANSI-colour assertion that fails on a colour-capable terminal; it is pre-existing and unrelated.

**Caveat worth stating:** I could not run the `kubeintellect-server` suite locally — its dependency set is much heavier than `kube-q`'s and I did not want to claim a green I had not seen. `app/cli.py` is 23 of the 62 sites, so that file is covered by review and by the AST rewrite being positional rather than textual, not by a local test run. CI is the real check there.


MSKazemi:
Thank you — this is the standard I want contributions to this repo held to, and I want to be
specific about why rather than just saying so.

**You widened the scope for the right reason.** #156 enumerated three files. You did 62 sites
across the whole CI-linted scope and said, in the PR description, exactly why: *"a gate that
exempts tests and probes is a gate with a carve-out, and 'the places nobody looks' is where
this class survives."* That is the correct read of how #136 reached a user. If I had written
the issue better it would have said that.

**And you stated a caveat that cost you something to state.** You could not run the server
suite, `app/cli.py` is 23 of the 62 sites, and you said so plainly instead of implying a green
you had not seen. I ran it: **1053 passed**, exactly the number your second commit put in
`AGENTS.md`.

Three other things worth naming: the edits are AST-positional rather than textual, so the
`**kwargs` and binary-mode cases are skipped for a reason rather than by luck; the gate is
stdlib-only and independent of ruff, which matters more than it looks because the `<0.16` pin
in `v4/pyproject.toml` is a standing blind spot this repo has already been bitten by; and
`test_encoding_gate.py` has real red cases plus a non-vacuity assertion, so the checker cannot
pass by scanning nothing.

## What I verified

All on your branch, before I touched anything:

| Gate | Result |
|---|---|
| `./scripts/check-text-encoding.py` | **308 files, clean**, exit 0 |
| `./scripts/check-syntax-warnings.py` | clean |
| `./scripts/check-file-modes.sh` | clean |
| `uv run ruff check …` | All checks passed |
| `uv run mypy …` | **0 errors, 172 source files** |
| server suite | **1053 passed** |
| `kube-q` suite | **317 passed** |

Your `test_export_to_file` failure did not reproduce here — my terminal is not colour-capable,
which is the other half of the same environment sensitivity #109 fixed. Nothing for you to do;
it confirms your read that it is pre-existing and unrelated.

## What I changed, and why

Two commits on top of yours, both `Co-authored-by:` you.

**1.** [`75c1f28`](https://github.com/MSKazemi/kubeintellect/commit/75c1f28cc375815ea200c4842f48c935ee5d852e)
— the gate matched on the call's **name**, so every `.open()` in the language was in scope.
Five lines, four flagged, none of which has a text encoding to name:

```
webbrowser.open(url)          -> TypeError: unexpected keyword 'encoding'
os.open(path, 0)              -> returns a file descriptor
zipfile.ZipFile(z).open(m)    -> binary handle
tarfile.open("a.tar")         -> no text encoding at this call
```

A contributor who followed the printed remediation would have broken their own code to turn CI
green. **Nothing in the tree calls any of these today** — I checked — so the gate was correct
on all 308 files and wrong only about the future, which is the part a gate is for.

Narrowed three ways, each with a red test: `<module>.open` on `webbrowser`/`os`/`zipfile`/
`tarfile`/`shelve`/`dbm` is out of scope (resolving one level of construction, so
`zipfile.ZipFile(z).open(m)` is seen as `zipfile` while `Path(p).open()` deliberately is not);
`gzip`/`bz2`/`lzma` default to **binary** unlike the builtin, so a call naming no mode is not
flagged but an explicit `gzip.open(p, "rt")` still is; and an encoding passed **positionally**
now satisfies the rule — `p.read_text("utf-8")` was flagged, and adding the keyword as
instructed is a duplicate-argument `TypeError`. That last one has an off-by-one of its own,
since the builtin `open` and `Path.open` put `encoding` at different indices; there is a red
test pinning both.

One hole rather than a false positive: `check_paths` caught `UnicodeDecodeError` alongside
`OSError` and skipped the file **without counting it**. A PEP 263 `# -*- coding: latin-1 -*-`
source would have passed the gate silently. That is a guard quietly ceasing to guard while CI
stays green, which this repo has been burned by before, so it is now reported with the file
named.

**2.** [`533ec1b`](https://github.com/MSKazemi/kubeintellect/commit/533ec1bce04d824375719ab4c47919d6007f3b5d)
— `CHANGELOG.md` entry crediting you, your README contributor row extended, and `test` + `tool`
added to your `.all-contributorsrc` entry. Recognition lands in the merge that carries the
work, not at release time.

Server suite is **1059** now (six tests added); `test_doc_claims` caught the `AGENTS.md` number
before I did, which is that gate doing its job.

## Re-run on the final tree

Because the tree I am merging is no longer the tree I reviewed:

```
check-text-encoding   308 files clean        ruff       All checks passed
check-syntax-warnings clean                  mypy       0 errors / 172 files
check-file-modes      clean                  server     1059 passed
                                             kube-q     317 passed
```

**All 9 required checks pass** on `533ec1b`.

## On PLW1514

Your reading is right, and I want to record the trade rather than leave it in a PR thread.
`PLW1514` is the smaller change *if it were reachable* — it is not in the linted scope today
and enabling it repo-wide lights up the frozen `v1/`–`v3/`. When the pin lifts (#75) and the
scope widens, deleting the script for a rule line is the right trade, and your header already
says so.

## One thing your PR found that it does not fix

`subprocess.run(..., text=True)` decodes with the locale encoding too — same bug class,
different call shape, **23 sites** in the non-frozen tree. The child processes here are
`kubectl`, `helm` and `git`, so the input is whatever users named their resources and whatever
a container wrote to its logs.

That is **#168**, and I deliberately did not fold it into this PR. It carries a real design
question rather than a mechanical sweep — whether `errors="replace"` belongs on the calls
reading cluster output, where a decode failure fails a whole diagnosis versus degrading one
glyph, against this project's rule that a plausible-looking wrong answer is the one thing it
cannot ship. It is written up with my inclination and the counter-argument. Yours if you want
it; no obligation, and it stays open either way.

Status: **approved, ready to merge**, holding only for the maintainer go on the merge itself.

MSKazemi (approved):
Approving. Feature verdict **WANT** — it closes #156, which I filed, and closes the class rather than the instances. Integration **SOUND**: it plugs into the gate family this repo already has (`check-syntax-warnings.py`, `check-file-modes.sh`), reuses the existing required `Syntax warnings` job rather than adding a check that would need a branch-protection change, and honours the two invariants that matter here — no dependency on ruff or the `<0.16` pin, and the frozen `v1/`–`v3/` untouched under ADR-001/002.

Evidence is in the review comment above: all 9 required checks green on `533ec1b`, and the full gate set re-run locally on the final tree after my two commits (encoding 308 files clean, syntax clean, modes clean, ruff clean, mypy 0/172, server 1059, kube-q 317).

MSKazemi:
Merged — `32e3740` on `main`.

**Where it landed.** `scripts/check-text-encoding.py` is now a repo-level gate: `make check-encoding`, and a step in the existing **Syntax warnings** CI job, so it runs on every PR from here on. All 62 call sites are on `main`. Your head commit `533ec1b` is in the history unchanged.

**What I changed on top**, and why — you should know exactly what is yours and what is mine:

- The gate flagged calls that take no `encoding` at all. `webbrowser`, `os`, `zipfile`, `tarfile`, `shelve` and `dbm` `.open()` are now exempt, and one level of construction is resolved so `zipfile.ZipFile(z).open(m)` is recognised while `Path(p).open()` deliberately stays in scope.
- `gzip`/`bz2`/`lzma` default to **binary**, so they are exempt unless an explicit text mode is given — `gzip.open(p, "rt")` is still caught.
- An encoding passed **positionally** is now accepted, at the right index for each of `read_text`, `write_text`, `Path.open` and builtin `open` (they differ; the off-by-one is a test of its own).
- A source file the gate cannot decode as UTF-8 was being **silently skipped**. That is now reported as a failure. This repo has already lost a guard exactly that way — a `ruff --fix` rewrote a canary and the suite still said green — so a gate that quietly stops guarding is the failure mode I care most about.
- Six tests for the above, all with hostile fixtures. Server suite 1053 → 1059.

**Verified before merge, on the merge commit itself:** encoding gate ✅, syntax gate ✅, file-modes gate ✅, `ruff check` 0, `mypy` 0 / 172 files, server suite **1059 passed**, kube-q suite **317 passed**.

**Recognition.** `.all-contributorsrc` now records `test` and `tool` for you alongside `code`, `doc`, `review`, and the README contributors table carries the full story of this one — the scope call included, because widening from three files to the whole linted scope, tests and probes included, is the reason the class is actually closed and not just quieter.

**One note for next time, purely mechanical:** fork PRs in this repo currently land as `BLOCKED` even when every required check is green, because CodeQL's default setup emits no check run on a fork head. That is not you and not your branch — I filed it as #170. It is why this took a manual merge rather than the button.

It ships in the next release; `CHANGELOG.md` under `## [Unreleased]` credits you by name.

If you want another one: #167 is the natural follow-up — nothing currently keeps `.all-contributorsrc` and the README contributors table in sync, and the same "gate the class, do not fix the instance" instinct applies exactly.

And if the project is useful to you, a ⭐ or a share helps more than it looks like it should — it is the whole reason anyone else finds it. Entirely optional, and it changes nothing about your standing here either way.

MSKazemi:
Naming an encoding on every text-mode read and write *and gating it* so it cannot regress is the version of that fix that actually holds. Most people would have changed the failing call site and moved on; you closed the class.

[IDKMesh](https://github.com/MSKazemi/idkmesh) is an earlier project of mine built almost entirely out of that idea — gates that make a property impossible to break, rather than merely fixed once. It is research on verified collaboration between humans and AI agents, and it is at the stage where the gates themselves are the interesting work. Today I found one of my own that had the failure mode you would predict: the local test gate cached a *failed* run as a pass, so after failing once it reported green for a tree that had never passed.

Honest about the stage: 1 star, no contributors but me. Newcomer issues are labelled [`good first issue`](https://github.com/MSKazemi/idkmesh/labels/good%20first%20issue), and [#400](https://github.com/MSKazemi/idkmesh/issues/400) — widening a link gate that silently ignores non-Markdown targets — is the closest to what you did here.

Thanks for the encoding work.

Issue #167, opened 2026-08-22
The two contributor rosters can drift apart, and did — nothing compares them


MSKazemi:
This repo keeps contributor recognition on **two** surfaces:

- `.all-contributorsrc` — the machine-readable one, per the [All Contributors](https://allcontributors.org/) spec
- the **Contributors** table in `README.md` — the one humans actually read

Nothing checks that they name the same people, and by 2026-08-22 they did not. The rc file
listed **5**; the README credited **8**; and [@Chris7717](https://github.com/Chris7717) was on
**neither**, despite a merged playbook (#114) and the report (#115) that found the `.gitignore`
glob keeping `.env.example` out of every clone.

That failed quietly for weeks. Prose credit is what people see, so the missing rc entries were
invisible; and a person missing from both surfaces is invisible everywhere.

## What to build

A gate in the family the repo already has — `scripts/check-syntax-warnings.py`,
`scripts/check-file-modes.sh`, `scripts/check-text-encoding.py`: **stdlib only, no virtualenv,
no network, independent of ruff and of the `<0.16` pin.**

It should assert that the set of logins in `.all-contributorsrc` equals the set of
`@handle`s in the README Contributors table. The comparison that found this is four lines:

```python
rc = {c["login"] for c in json.loads(Path(".all-contributorsrc").read_text(encoding="utf-8"))["contributors"]}
readme = set(re.findall(r"^\| \*\*\[@([^\]]+)\]", Path("README.md").read_text(encoding="utf-8"), re.M))
assert rc == readme
```

Please make the failure name *which* handles are on which side — a gate that just says
"they differ" sends the next person hunting through a 9-row table.

## Done when

- [ ] `scripts/check-contributor-roster.py` exists, stdlib only, exits 1 naming the missing handles on each side
- [ ] wired into `make check-roster` and the existing `Syntax warnings` CI job (**do not rename that job** — branch protection matches required checks by name, so a rename blocks every open PR until the settings are updated)
- [ ] a test alongside `v4/tests/test_encoding_gate.py` with a **red case**: a fixture missing from one side must fail, or the gate is worth nothing
- [ ] `AGENTS.md`'s server-suite count updated — `test_doc_claims.py` will tell you the number

## Notes

The regex above is deliberately anchored to the exact table row shape. If you would rather
parse the table properly, that is welcome — just keep it stdlib.

Context: the roster gap itself is fixed in the PR that links here; this issue is only about
keeping it fixed.

Issue #154, opened 2026-08-21
Count-drift gates run per-PR and cannot see cross-PR drift


MSKazemi:
Merging #133 and #138 turned `main` red, and neither PR was at fault.

Both were green on their own head SHA. Both were branched 31 commits back, and both added tests. `tests/test_doc_claims.py` compares a hard-coded server-suite count in `AGENTS.md` against a live collect — **neither PR's CI run ever saw the other's tests**, so each was individually consistent and the merge of the two was not. 1040 claimed, 1047 collected. `Tests (server)` and `Tests (server · py3.13)` are both required checks, so `main` went red and stayed blocking until #152.

This is the same count-drift family as the earlier playbook-count problem, but with a new trigger: **it is not reachable from any single PR**. A per-PR gate structurally cannot see a number that only two merges *together* move, so no amount of care on the contributor side prevents it.

## Options

1. **Run the doc-claims gate on `main` after merge** — catches it, but only after `main` is already red. Cheap, and at least the signal is immediate and unambiguous.
2. **Have the PR job collect against the merge result** rather than the PR head. GitHub already builds a merge commit for `pull_request` events; collecting there would have caught this *before* either merge. Better, and probably the real fix.
3. **Stop hard-coding the count.** The number in `AGENTS.md` is documentation of a fact the test suite already knows. A generated line, or a tolerance, removes the whole class — at the cost of losing a genuine tripwire that has caught real drift before.

Option 2 plus keeping the exact count is my current preference, but this is worth deciding rather than patching again.

## Also worth noting

The same structural blind spot applies to every other hard-coded count the doc-claims checker guards (playbooks, detectors, flags, providers). None of them are safe from a two-PR interaction, only from a one-PR one.

MSKazemi:
**It happened again on 2026-09-09 — and this time the root cause is narrower than option 2 assumes, which makes the fix a settings change rather than a workflow change.**

## Fresh reproduction (a near-miss, not another red `main`)

Two PRs in flight yesterday, both adding tests, both branched from the same `main`:

| PR | what `check_doc_claims.py` computed **in isolation** |
|---|---|
| #184 (parallel tool isolation, +5 tests) | `5510` |
| #187 (DNS playbook, +5 tests) | `5510` |
| **actual total once both merged** | **5515** |

Exactly the shape described here: each PR was individually consistent, neither CI run could see the other's tests, and the merge of the two was not consistent.

[...]

PR #172, opened 2026-08-23
fix(encoding): subprocess.run(..., text=True) decoded with the locale encoding too (#168)

### LINKED ISSUE #168: subprocess.run(..., text=True) decodes with the locale encoding too — 23 sites the new encoding gate cannot see
Follow-up from #161, which closed the `read_text()`/`write_text()`/`open()` half of the
#136 bug class and left this half standing.

`subprocess.run(..., text=True)` **also** decodes with the platform default when no
`encoding=` is given. On Windows (CP1252/CP936) or under the POSIX `C` locale, any non-ASCII
byte in the child process's output raises `UnicodeDecodeError` — the same failure, the same
platforms, a different call shape.

This is not theoretical for this project. The child processes are `kubectl`, `helm` and `git`,
and their output carries whatever users named their resources and whatever a container wrote
to its logs. A pod annotation with a non-ASCII character is enough.

## Scope

23 call sites in the CI-linted tree (excluding the frozen `v1/`–`v3/`):

```console
$ grep -rn --include=*.py "text=True" --exclude-dir=v1 --exclude-dir=v2 --exclude-dir=v3 . \
    | grep -v "encoding=" | wc -l
23
```

They include `scripts/check-syntax-warnings.py` and `scripts/check-text-encoding.py`
themselves, `v4/scripts/aci_sandbox_probe.py`, `opsmembench_live_probe.py`, `fix_pr_probe.py`
and `aci_token_ab.py`. The kubectl/helm tool paths are the ones that matter to a user.

`scripts/check-text-encoding.py` deliberately does **not** cover this — its header says so —
because the receiver is a `subprocess` call rather than a file open, and lumping the two into
one checker would make both harder to read.

## What to do

1. Add `encoding="utf-8"` to every `subprocess.run` / `Popen` / `check_output` call that sets
   `text=True` (or `universal_newlines=True`) in the non-frozen tree.
2. Decide whether `errors="replace"` belongs on the ones reading **cluster** output. This is
   the interesting design question and worth arguing in the PR rather than assuming: a
   `UnicodeDecodeError` on `kubectl logs` fails a diagnosis outright, where `errors="replace"`
   degrades one glyph and still returns the logs. The counter-argument is that silently
   replacing bytes is how a wrong answer looks right — which this project treats as the one
   failure mode it cannot ship. **My inclination is `errors="replace"` on log/output reads and
   strict elsewhere**, but I would rather see the case made.
3. Extend the gate to cover the shape, once (2) is settled, so the class cannot come back.

## Done when

- [ ] no `text=True` without an `encoding=` outside `v1/`–`v3/`
- [ ] the `errors=` decision is recorded in the PR description with its reasoning
- [ ] `scripts/check-text-encoding.py` (or a sibling) catches the shape, with a red test
- [ ] a regression test that a non-UTF-8 byte in child output does not blow up the caller

Related: #136, #156, #161.

MSKazemi:
Closes #168.

`subprocess.run(..., text=True)` decodes the child's pipes with the platform default,
exactly as `read_text()` decodes a file with it. #161 closed the file half of #136/#156
and left this one standing. The children here are `kubectl`, `helm` and `git`, so the
bytes are whatever a user named a resource and whatever a container wrote to its logs —
on Windows (CP1252/CP936) or under the POSIX `C` locale, one non-ASCII byte is enough.

### Scope — 25, not 23

The issue counted 23. On `main` today it is **25 call sites in 12 files**: #167/#171 added
one (`check-contributor-roster.py`) and #161 grew two of its own. The issue's own grep now
prints 26, one of which is the `NOT COVERED` comment in the gate header.

### The `errors=` decision, which #168 asked to be argued rather than assumed

The inclination in the issue was `errors="replace"` on cluster output and strict elsewhere.
That is very close, but *cluster-vs-not* is not the line that holds. The line is
**content vs identifier**:

> `errors="replace"` where the decoded bytes are free-form **content** a human or the model
> will read. Strict where they are an **identifier**, a **path**, or a tool's own
> **structured output**.

Three things make the replace side safe here, and they are specific to these call sites:

1. **The output is already lossy.** Every content site truncates (`[:_SNAPSHOT_MAX_CHARS]`,
   `max_chars=4000`) and redacts (`redact_secrets`) before anything reads it. `errors="replace"`
   is one more lossy transform of the same kind — it does not cross a new line.
2. **U+FFFD is not silent, and the status quo is.** The objection to replacement is that
   silently substituting bytes is how a wrong answer looks right. But `�` is visible and
   greppable. What is actually silent today is *strict*: three of the content sites wrap the
   call in a broad `except Exception` — `_capture_rollback_point` (`continue`),
   `cluster_id._kubectl` (`return ""`), `context_fetcher._run_kubectl_snapshot`
   (`return "(unavailable: …)"`). There, one unreadable glyph currently means **the rollback
   point was never armed**, with no marker at all.
3. **Two failures are severe.** `run_kubectl` catches only `FileNotFoundError` and
   `TimeoutExpired`, so a `UnicodeDecodeError` escapes into the agent loop.
   `api/v1/endpoints/namespaces.py` has no `try` at all — it is a 500 on a public endpoint.

And what makes the strict side *correct* rather than merely conservative: every identifier read
here is **ASCII by specification**. Namespace and context names are RFC-1123 constrained;
`which` returns a path; `systemctl is-active` returns a literal. A `�` in any of those corrupts
a **key** — it would fork a cluster identity, or name a namespace that does not exist — where
strict decoding can only fire on something genuinely wrong. `git ls-files` is the sharpest case:
replacing a byte there yields a path that cannot be opened, which is a worse and far more
confusing failure than the decode error.

One site sits on the boundary and went **strict deliberately**: `v4/scripts/check_doc_claims.py`
regexes a count out of pytest output. It reads free-form text, but it is a *gate*, and a gate
that dies loudly on a stray byte is behaving correctly.

Two sites (`cli.py:1636`, `cli.py:1649`) pass `text=True` with no pipes, where it decodes
nothing at all today. They were annotated rather than deleted: the moment someone adds
`capture_output=True`, they become locale-dependent silently.

**Split: 9 `errors="replace"` / 16 strict.**

| `errors="replace"` — content | strict — identifier / path / tool output |
|---|---|
| `kubectl_tool` rollback `get -o yaml` | `namespaces.py`, `cluster_id.py` |
| `kubectl_tool.run_kubectl` (logs/describe/get) | `cli.py` ×8 |
| `helm_tool.run_helm` | `check_doc_claims.py` |
| `context_fetcher` pods + events | `check-syntax-warnings.py` ×2 |
| `aci/gitops` git/gh reason string | `check-contributor-roster.py` |
| the four `v4/scripts/*_probe`-style scripts | `check-text-encoding.py` ×2 |

### The gate

`scripts/check-text-encoding.py` was **extended in place** rather than given a sibling — its
header already named this shape as "NOT COVERED, tracked separately", so that header is now the
documentation of the `errors=` convention too.

Detection matches on the **pairing** of a `text=`/`universal_newlines=` keyword with a call named
`run`/`Popen`/`check_output`/`call`/`check_call`. Matching the name alone would flag every
`runner.run(...)` in the tree; requiring a `subprocess.` receiver would miss
`from subprocess import run`. A non-literal `text=flag` is flagged (it may be `True` at runtime);
an explicit `text=False` is not.

It rides in the existing **Syntax warnings** job. Branch protection matches required checks by
name, so a new job would be one `main` does not require — the same reasoning as #167.

### Evidence

| Check | Result |
|---|---|
| AST sweep of the whole tracked tree | **0** subprocess text-mode calls without `encoding=`, 310 files |
| extended gate, patched tree | `exit 0` |
| extended gate, 5 **pre-fix** files restored from `main` | `exit 1`, **14** offenders — the red case |
| `check-syntax-warnings` / `check-contributor-roster` / `check-file-modes` | `exit 0` |
| `ruff check`, all 19 changed files | `All checks passed!` |
| `mypy` (app + ki-protocol + kube_q) | **Success: no issues found in 172 source files** |
| `pytest tests/` **before** | 13 failed, 1055 passed |
| `pytest tests/` **after** | 13 failed, **1066** passed — same 13, **zero new** |

Those 13 are pre-existing on pristine `main` in that local environment (GPU collector, prometheus
trends, budget gate, version axes). CI is green on `main`, so they are environmental; they were
measured only to establish the baseline and are **not** diagnosed here.

### Done-when, against #168

- [x] no `text=True` without an `encoding=` outside `v1/`–`v3/`
- [x] the `errors=` decision recorded with its reasoning (above)
- [x] the gate catches the shape, with red tests
- [x] a regression test that a non-UTF-8 byte in child output does not blow up the caller —
      `v4/tests/test_child_output_that_is_not_utf8.py` spawns a real child emitting a raw `0x80`
      and asserts **both** directions (replace survives with a visible marker; strict raises),
      then AST-pins `errors=` onto the four cluster-output readers by name, so dropping it from
      `run_kubectl` again cannot pass green

### Note for the merger — this PR is itself an instance of #154

It moves the server-suite count `1068 → 1079` in `AGENTS.md`. If anything else that adds tests
merges first, this goes red on `main` rather than here, because branch protection has
`required_status_checks.strict = false`. Re-run before merging, or merge this one first.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -59,7 +59,7 @@ uv run ruff check packages/kubeintellect-server/app/ packages/ki-protocol/
 # 2. Types — the workspace is at ZERO errors; keep it there.
 uv run mypy packages/kubeintellect-server/app packages/ki-protocol packages/kube-q/kube_q
 
-# 3. Server suite (1068 tests)
+# 3. Server suite (1079 tests)
 uv run python -m pytest tests/ -q
 
 # 4. kq CLI suite (317 tests)
@@ -79,7 +79,8 @@ make fix-modes            # corrects any violation in place
 # 6. Syntax — every tracked .py outside v1-v3 compiles with no SyntaxWarning.
 make check-syntax         # or: ./scripts/check-syntax-warnings.py
 
-# 7. Encoding — every text-mode read/write names an encoding (#136/#156).
+# 7. Encoding — every text-mode read, write and subprocess call names an
+#    encoding (#136/#156/#161, and the subprocess half #168).
 make check-encoding       # or: ./scripts/check-text-encoding.py
 
 # 8. Roster — .all-contributorsrc and the README table name the same people (#167).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,41 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Fixed
+- **`subprocess.run(..., text=True)` decoded with the locale encoding too** (#168) — the half of
+  the #136/#156 class that #161 left standing. The children here are `kubectl`, `helm` and `git`,
+  whose output carries whatever a user named a resource and whatever a container wrote to its
+  logs, so on Windows or under the POSIX `C` locale a single non-ASCII byte was enough.
+
+  25 call sites named an encoding. The failure was not uniform, which is why the fix is not
+  either: `run_kubectl` catches only `FileNotFoundError`/`TimeoutExpired`, so the decode error
+  escaped into the agent loop; `GET /namespaces` had no handler at all and would have returned
+  a 500; and three sites — the rollback-point snapshot, `cluster_id`, the context fetcher's pod
+  and event snapshot — sit inside a broad `except Exception`, where the same byte silently meant
+  *the safety feature did not run*.
+
+  `errors="replace"` was applied to the 9 sites that read free-form **content** a human or the
+  model reads, and withheld from the 16 that read an **identifier, a path, or a tool's own
+  structured output**. The line is content-vs-identifier rather than cluster-vs-not: a `U+FFFD`
+  in `kubectl logs` blemishes prose that was already truncated and redacted before anyone saw
+  it, while the same character in a namespace name, a cluster ID or a `git ls-files` path
+  corrupts a key — and every one of those values is ASCII by specification, so strict decoding
+  there can only fire on something genuinely wrong.
+
+  `scripts/check-text-encoding.py` was extended to the shape rather than given a sibling, since
+  its header already named this as the tracked-separately half; it now documents the `errors=`
+  convention as well. Detection matches the pairing of a `text=`/`universal_newlines=` keyword
+  with a call named `run`/`Popen`/`check_output`/`call`/`check_call` — the name alone would flag
+  every `runner.run(...)` in the tree, and requiring a `subprocess.` receiver would miss
+  `from subprocess import run`. It rides in the existing **Syntax warnings** job, so no required
+  check changes name.
+
+  Eleven tests: five red/green cases on the gate, and a new
+  `v4/tests/test_child_output_that_is_not_utf8.py` that spawns a real child emitting a raw
+  `0x80` and asserts both directions — replace survives with a visible marker, strict raises —
+  then AST-pins `errors=` onto the four cluster-output readers by name, so dropping it from
+  `run_kubectl` again cannot pass green.
+
 ### Added
 - **A gate now keeps the two contributor rosters in sync** (#167).
   `.all-contributorsrc` and the **Contributors** table in `README.md` are both hand-maintained,
--- a/scripts/check-contributor-roster.py
+++ b/scripts/check-contributor-roster.py
@@ -64,6 +64,7 @@ def repo_root() -> str:
         ["git", "rev-parse", "--show-toplevel"],  # noqa: S607
         capture_output=True,
         text=True,
+        encoding="utf-8",
         check=True,
     ).stdout.strip()
 
--- a/scripts/check-syntax-warnings.py
+++ b/scripts/check-syntax-warnings.py
@@ -62,6 +62,7 @@ def repo_root() -> str:
         capture_output=True,
         check=True,
         text=True,
+        encoding="utf-8",
     ).stdout.strip()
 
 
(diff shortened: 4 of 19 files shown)


MSKazemi:
Marking this clearly for anyone who lands on #168: **this draft is stale and should not be read as "the issue is taken".**

Opened 23 August, now conflicting, and I am not actively working it. The underlying problem has meanwhile grown from **23 to 53** sites — re-measured today, with a reproduction posted on #168.

#168 is **open and available**. Whoever takes it may lift anything useful from this branch or ignore it entirely. I will close this once the issue is claimed, or finish it myself if nobody wants it — but it should not sit there discouraging someone in the meantime.

MSKazemi:
**Status update: [#168](https://github.com/MSKazemi/kubeintellect/issues/168) is now claimed — by [@1cbyc](https://github.com/1cbyc) — so my note above ("open and available") is out of date.**

Leaving this draft open on purpose, because it is now a reference rather than a claim. @1cbyc, per my earlier note: lift anything useful from this branch, and ignore anything that disagrees with the scope we agreed on #168. It is from 23 August, it conflicts with `main`, and the site count has since gone 23 → 53, so treat it as a sketch of the shape and not as a head start you are obliged to use.

One thing in here that is still worth reading rather than re-deriving: the call sites are **not** uniform, and a bulk rewrite is the wrong instinct — which is exactly the distinction @1cbyc drew unprompted when claiming it (strict `utf-8` for identifiers and structured command output, `errors="replace"` only where the purpose is displaying free-form logs). That framing is better than the one recorded in the issue body, and it is the version we are going with.

I will close this branch once #168 lands.

MSKazemi:
Closing as promised — #168 landed via #213 (@1cbyc, merged as `b81358c`), which took the same content-vs-identifier line this draft sketched and shipped it correctly, plus three sites I finished on top in review. Nothing here is needed as a separate branch anymore. Thank you again for leaving it as a reference rather than letting it sit as a competing claim.

Issue #173, opened 2026-08-27
Coordinator emits unresolved kubectl placeholders that are guaranteed to fail the command safety gate


MSKazemi:
### Summary

A real `kq` session hit a deterministic planning failure while trying to solve several cluster problems. The coordinator emitted an unresolved metavariable as an actual tool argument:

```text
You: solve ths issues
⚙ run_kubectl → Running: kubectl describe pod ml-scorer-5bc9565b96-5tskr -n shop
⚙ run_kubectl → Running: kubectl describe pod payments-api-56f659864c-28s2w -n shop
⚙ run_kubectl → Running: kubectl get nodes -o wide
⚙ run_kubectl → Running: kubectl describe node <node-name>
...
Error: LLM error: Command contains disallowed shell characters ...
```

`run_kubectl` is correct to reject `<` and `>`: they are shell metacharacters and the runner intentionally blocks them. The bug is upstream: a placeholder that exists only as planning notation reached the execution surface.

### Why this is a coordinator bug, not a sanitizer bug

The v4 coordinator prompt already says dependent calls must be sequential, but its parallel-execution guidance includes a `describe node` example without establishing a concrete node first. In this trace the model correctly knew it needed `kubectl get nodes`, but incorrectly scheduled `kubectl describe node <node-name>` in the same batch instead of waiting for the node name.

Do **not** fix this by weakening `_SHELL_METACHAR` in `app/tools/kubectl_tool.py`. The safety gate should continue rejecting `<` / `>`.

### Expected behavior

If a command needs an identifier that is not already present in the cluster snapshot or conversation state, the coordinator should:

1. fetch the identifier first (`kubectl get nodes -o wide`),
2. wait for that result,
3. emit a second tool call using the concrete node name.

No metavariable such as `<node-name>`, `<pod>`, `<namespace>`, `$NODE`, `{node}`, or similar planning placeholder should ever be passed to `run_kubectl`.

### Suggested scope

- Tighten the coordinator system prompt around dependent calls and replace the ambiguous parallel `describe node` example.
- Add a pre-execution/model-output guard that detects unresolved metavariables in tool arguments and routes them back for replanning rather than invoking the tool.
- Prefer resolving names from the existing snapshot before doing another discovery call when possible.

### Acceptance criteria

- [ ] `run_kubectl` continues to reject `<` and `>` as shell metacharacters.
- [ ] The coordinator never invokes `run_kubectl` with an unresolved placeholder/metavariable.
[...]

Issue #170, opened 2026-08-22
Fork PRs can never reach a mergeable state — CodeQL default setup produces no results on them


MSKazemi:
Found while merging #161, and it affects **every** outside contribution.

## What happens

#161 is a fork PR. On its head SHA `533ec1b`:

- all **9 required status checks pass**
- `reviewDecision` is `APPROVED`
- `statusCheckRollup` is `SUCCESS`
- there are **0** unresolved review threads

and `mergeStateStatus` is still **`BLOCKED`**:

```console
$ gh pr merge 161 --squash
X Pull request #161 is not mergeable: the base branch policy prohibits the merge.
```

The cause is code scanning. CodeQL here is **default setup**, not a workflow file:

```console
$ gh api repos/MSKazemi/kubeintellect/code-scanning/default-setup
{"state":"configured","languages":["actions","javascript","javascript-typescript","python","typescript"],"query_suite":"default","schedule":"weekly"}

$ gh api repos/MSKazemi/kubeintellect/commits/533ec1b.../check-runs | grep -i codeql
(nothing)
```

Zero CodeQL check runs on the fork head. On a same-repo branch (#169) all three
`Analyze (…)` jobs run and pass. The branch requires code scanning results; a fork PR can
never produce them; so the merge box never clears.

## Why it matters more than it looks

Every outside contribution has to be merged with `--admin`, which means **the routine path for
merging a stranger's PR is to bypass branch protection.** That is a bad habit to build into a
project whose growth depends on outside contributions — the one class of PR where the scan is
most worth having is the one class that cannot run it, and the workaround trains everyone to
reach for the override.

It is also invisible: nothing says "blocked because code scanning did not run". It presents as
a green PR that will not merge.

## Options, none of them free

1. **Move CodeQL from default setup to an advanced-setup workflow file** (`.github/workflows/codeql.yml`)
   triggered on `pull_request`. Workflows on fork PRs run with a read-only token and no secrets,
   which is fine for CodeQL. This is the option that actually gets fork PRs scanned. Cost: the
   workflow becomes ours to maintain, and moving off default setup is a one-way-ish door.
2. **Drop code scanning from the required checks** and rely on the scan of `main` after merge
   plus the weekly schedule. Cheapest, and honest as long as it is written down — but it means
   nothing scans a contributor's code before it lands.
3. **Keep it and accept `--admin` per fork PR**, documented in `TRIAGE.md` so it is a decision
   rather than a habit.

[...]

PR #183, opened 2026-09-06
fix(agent): isolate failures in parallel tool batches

### LINKED ISSUE #174: One invalid tool call aborts an entire parallel Kubernetes investigation batch
### Summary

A real `kq` remediation attempt emitted several independent read-only `run_kubectl` calls in one batch. Most commands were valid, but one invalid command caused the entire investigation to terminate with an LLM/tool error instead of preserving the successful results and letting the coordinator continue.

Observed session:

```text
You: solve ths issues
⚙ run_kubectl → Running: kubectl describe pod ml-scorer-5bc9565b96-5tskr -n shop
⚙ run_kubectl → Running: kubectl describe pod payments-api-56f659864c-28s2w -n shop
⚙ run_kubectl → Running: kubectl get nodes -o wide
⚙ run_kubectl → Running: kubectl describe node <node-name>
⚙ run_kubectl → Running: kubectl describe pod checkout-5f9f68c4cb-cfgs4 -n shop
⚙ run_kubectl → Running: kubectl get events -n shop --sort-by=.lastTimestamp
⚙ run_kubectl → Running: kubectl describe pod report-worker-d6b7d44f9-rkfh6 -n shop
Error: LLM error: Command contains disallowed shell characters ...
```

The bad call was `kubectl describe node <node-name>`. That call should fail, but its failure should not discard or suppress the results of the other independent reads.

### User impact

This is especially damaging on broad incident-remediation requests, where the coordinator intentionally fans out many independent reads. A single malformed or unsupported call can turn a mostly successful investigation into a total failure and forces the user to restart the whole request.

### Expected behavior

For a parallel batch of independent reads:

- each tool call should produce its own `ToolMessage` success or structured error,
- successful results should remain available to the coordinator,
- one call raising `ValueError` / validation failure should be converted to a tool-result error rather than escaping the graph,
- the coordinator should synthesize from successful calls and, if useful, re-plan only the failed branch.

This should remain distinct from HITL interrupts: mutation approval must still interrupt according to the existing safety model.

### Suggested scope

- Inspect the coordinator/cortex tool execution path for exception aggregation behavior.
- Catch ordinary tool invocation errors at the per-call boundary and return a structured failure `ToolMessage` containing command + reason.
- Preserve `GraphInterrupt` semantics for HITL; do not blanket-catch it as an ordinary tool failure.
- Add observability so a batch can report e.g. `6 succeeded, 1 failed` rather than surfacing a generic `LLM error`.

### Acceptance criteria

- [ ] A parallel batch containing six valid read calls and one invalid `run_kubectl` call returns six successful tool results plus one structured failure.
- [ ] The coordinator receives all seven results and can continue/synthesize instead of the request terminating.
- [ ] The user sees which command failed and why.
- [ ] HITL / `GraphInterrupt` behavior for mutations is unchanged.
- [ ] Regression test covers a `ValueError` from one `run_kubectl` call inside a multi-call response.

### Related

Issue #173 covers the upstream cause of the specific invalid placeholder command. This issue is about fault isolation: even if the model emits one bad call in the future, independent work should not be lost.

be-student:
## What & why

A malformed tool call currently raises out of LangGraph's parallel `ToolNode`, discarding successful investigation results from the same batch. This adds a shared fault-isolation boundary for coordinator and specialist ReAct loops: ordinary failures become redacted per-call error results, while `GraphInterrupt` still escapes for HITL approval.

Closes #174

## Type of change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [x] 📖 Docs
- [ ] 🧹 Refactor / chore
- [ ] ⚡ Performance
- [x] 🧪 Tests

## Scope

## Checklist

- [ ] `uv run pytest` passes locally
- [x] Docs updated if behavior/CLI/flags changed

## Notes for reviewers

- The regression builds a real compiled LangGraph with seven parallel calls and proves six valid results survive one malformed call.
- A separate test proves `GraphInterrupt` is not converted into a tool error; another proves command and exception credentials are redacted.
- Focused agent tests: 71 passed.
- Full server suite: 5,455 passed, 23 skipped, 29 host/pre-existing failures. The failures require a missing monospace font, rely on GNU `chmod --` behavior unavailable on macOS, or flag floating GitHub Action references already present on `main`; none touch the changed paths.
- Exact workspace Ruff, mypy, doc-claims, file-mode, syntax, encoding, and contributor-roster gates passed.

Substantial AI assistance was used to inspect the LangGraph execution boundary, implement the change, and develop the tests. I reviewed the resulting diff and ran the validation listed above.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -257,6 +257,12 @@ A change that violates any of these will be rejected regardless of test results.
    behavioural test while silently receiving `None`. A behavioural test cannot catch that;
    only asserting the annotation can.
 
+7. **One read-tool failure must not abort a parallel batch.** The default coordinator and
+   specialist ReAct loops must pass tools through
+   `app.agent.tool_execution.fault_isolated_tool_node`. Ordinary invocation errors become
+   per-call error `ToolMessage`s that include the failed command and reason, while
+   `GraphInterrupt` must still escape unchanged for HITL mutation approval.
+
 ## Testing expectations
 
 - Every behavioral change needs a test that **fails before your change and passes after.**
--- a/v4/packages/kubeintellect-server/app/agent/nodes/coordinator.py
+++ b/v4/packages/kubeintellect-server/app/agent/nodes/coordinator.py
@@ -16,6 +16,7 @@ from langchain_core.runnables import RunnableConfig
 from langgraph.errors import GraphInterrupt, GraphRecursionError
 
 from app.agent.state import AgentState, PlanStep
+from app.agent.tool_execution import fault_isolated_tool_node
 from app.core.config import settings
 from app.core.llm import get_coordinator_llm
 from app.streaming.emitter import PlanEvent, StatusEvent, emit
@@ -853,7 +854,7 @@ async def _direct_answer(state: AgentState, config: Optional[RunnableConfig] = N
         system_parts.append(_PROACTIVE_FIX_BLOCK)
 
     llm = get_coordinator_llm()
-    agent = create_react_agent(llm, tools=ALL_TOOLS)
+    agent = create_react_agent(llm, tools=fault_isolated_tool_node(ALL_TOOLS))
 
     history, history_summary = _trim_session_messages(list(state["messages"]))
     if history_summary:
--- a/v4/packages/kubeintellect-server/app/agent/nodes/subagent.py
+++ b/v4/packages/kubeintellect-server/app/agent/nodes/subagent.py
@@ -7,6 +7,7 @@ from langchain_core.messages import SystemMessage
 from langgraph.prebuilt import create_react_agent
 
 from app.agent.state import AgentFinding, SubagentInput
+from app.agent.tool_execution import fault_isolated_tool_node
 from app.core.llm import get_subagent_llm
 from app.tools.registry import ALL_TOOLS
 from app.utils.logger import get_logger
@@ -91,7 +92,7 @@ async def run_subagent(payload: SubagentInput) -> AgentFinding:
     system_prompt = "\n".join(system_parts)
 
     llm = get_subagent_llm()
-    agent = create_react_agent(llm, tools=ALL_TOOLS)
+    agent = create_react_agent(llm, tools=fault_isolated_tool_node(ALL_TOOLS))
 
     input_messages = [SystemMessage(content=system_prompt)] + list(messages)
 
--- a/v4/packages/kubeintellect-server/app/agent/tool_execution.py
+++ b/v4/packages/kubeintellect-server/app/agent/tool_execution.py
@@ -1,0 +1,49 @@
+"""Shared fault-isolation boundary for ReAct tool batches."""
+from __future__ import annotations
+
+from collections.abc import Awaitable, Callable, Sequence
+
+from langchain_core.messages import ToolMessage
+from langchain_core.tools import BaseTool
+from langgraph.errors import GraphInterrupt
+from langgraph.prebuilt import ToolNode
+from langgraph.prebuilt.tool_node import ToolCallRequest
+from langgraph.types import Command
+
+from app.utils.logger import get_logger
+from app.utils.redact import redact_secrets
+
+logger = get_logger(__name__)
+
+
+async def _isolate_tool_failure(
+    request: ToolCallRequest,
+    execute: Callable[[ToolCallRequest], Awaitable[ToolMessage | Command]],
+) -> ToolMessage | Command:
+    """Convert one ordinary failure to a result without weakening HITL."""
+    try:
+        return await execute(request)
+    except GraphInterrupt:
+        raise
+    except Exception as exc:
+        call = request.tool_call
+        name = call.get("name", "")
+        command = (call.get("args") or {}).get("command")
+        safe_command = redact_secrets(command, max_chars=500) if isinstance(command, str) else ""
+        invocation = f" command={safe_command!r}" if safe_command else ""
+        reason = redact_secrets(str(exc), max_chars=500)
+        logger.warning(
+            f"tool call failed independently: {name}{invocation}: {reason}",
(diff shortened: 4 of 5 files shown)


MSKazemi:
Thank you, @be-student — this is an excellent first contribution, and it is **accepted**. 🎉

I want to be specific about why, because you got several things right that this class of patch usually gets wrong.

## What I verified

I did not take the description on trust; I reproduced the bug and then re-ran every gate on your head SHA (`01383c5`, Python 3.13).

**The premise is exactly right.** I built the control — stock `ToolNode`, your seven-command batch — and confirmed the abort:

```
CONTROL: batch ABORTED -> ValueError: Command contains disallowed shell characters
         <-- 0 of 6 good results survive
```

And the reason is precisely what you said. LangGraph's default handler is:

```python
def _default_handle_tool_errors(e: Exception) -> str:
    if isinstance(e, ToolInvocationError):
        return e.message
    raise e
```

It covers argument-binding failures and **re-raises everything else** — so any exception from inside a tool body takes the batch down. On an incident-response tool, returning an error where six correct answers existed is close to the worst outcome we can produce, which is why #174 was filed the way it was.

**Things you did that I specifically want to name:**

- You found the right seam. `awrap_tool_call` is the correct extension point; wrapping a `try` around the graph or setting `handle_tool_errors=True` would both have worked in the demo and been wrong — the first loses per-call granularity, the second swallows the HITL interrupt.
- **You kept the interrupt escaping.** Safety invariant #2 is the one thing in this repo I will reject a PR over, and you protected it without being asked and wrote a test for it.
- You put both the command and the failure reason through `redact_secrets` before returning them to the model. Nothing in the issue asked for that. It is the detail most fault-isolation patches miss.
- Your regression builds a **real compiled LangGraph**, not a mock. It fails against unpatched code for the right reason.
- And you disclosed the AI assistance, which is all we ask — it tells me where to look, nothing more.

## What I changed on top

Landed as a second commit in #184; your commit is carried **unmodified, with your authorship intact**.

**1. The boundary now re-raises `GraphBubbleUp` rather than `GraphInterrupt`.** This is the one real finding. `GraphInterrupt` is not the only signal that must escape — LangGraph marks all of them with a shared base class:

```
GraphBubbleUp.__subclasses__() -> [GraphDrained, GraphInterrupt, ParentCommand]
```

`ParentCommand` is what LangGraph raises when a tool routes the parent graph. Catching only `GraphInterrupt` meant the other two were isolated as if they were ordinary failures — a parent-graph route would come back as an error `ToolMessage` and the routing would be lost. **Nothing in the tree raises `ParentCommand` today, so this was latent, not live** — and latent is exactly how it would have reached production, because no behavioural test over today's tools can see it. That is the same shape as invariant #6, which is why it is now written down as invariant #7 rather than left in a commit message.

Two tests cover it, both failing against the narrower boundary:

```
### REVERTED to GraphInterrupt-only:  2 failed, 3 passed
### With GraphBubbleUp:               5 passed
```

**2. Minor housekeeping**, none of it a criticism — these are gates you had no way to predict from outside:

- The server-suite count in `AGENTS.md` is itself gated (`check_doc_claims.py`), so adding tests requires updating it: 5505 → 5510. Your PR would have gone red on a check that only runs in CI.
- I made the two new files `ruff format` clean so the repo's format-drift number stays at 127 instead of growing to 128. New files may as well be clean even though the tree overall is not.

**On your 29 reported failures:** you read those correctly — they are environmental (missing font, GNU `chmod`, floating action refs). For the record, on my machine only **one** survives, `test_every_action_is_pinned_to_a_commit_sha`, and I confirmed it fails on `origin/main` untouched. It is ours, not yours.

## Gates on the merged result

| Gate | Result |
|---|---|
| `ruff check` | All checks passed |
| `mypy` | no issues in 193 source files |
| Server suite | 5486 passed, 23 skipped |
| kq suite | 749 passed |
| doc-claims / modes / syntax / encoding / roster | all pass |

## Credit

You are now in [`CHANGELOG.md`](https://github.com/MSKazemi/kubeintellect/blob/main/CHANGELOG.md), the [`README` contributors table](https://github.com/MSKazemi/kubeintellect#contributors) and `.all-contributorsrc`, and you will be named in the next release notes. Because your commit is carried with its original authorship, it also lands on your GitHub contributions graph.

I am closing this PR in favour of #184 purely for the mechanical reason in #170 — fork PRs here cannot reach a mergeable state because CodeQL's default setup returns no results on them. **It is a merge, not a rejection**, and #184 says so at the top.

---

If you would like more of this: #173 is the sibling of the issue you just fixed — the coordinator emits unresolved `kubectl` placeholders like `<node-name>` that are *guaranteed* to fail the safety gate. Your change means those no longer destroy the batch; #173 is about not emitting them in the first place. You now know that code path better than anyone. It is yours if you want it, and I would happily assign it.

And one entirely optional thing — please read it as an invitation, not a condition, since the work is already merged either way. This fix is the kind that is invisible from the outside but matters a lot to anyone running the thing during an incident, and I would rather it be visible than not. If you think it is worth other people's attention, a ⭐ on the repo, or a short post about the `awrap_tool_call` seam and why `GraphBubbleUp` is the right catch — on LinkedIn, X, or wherever you write — would genuinely help. That detail is not well documented anywhere and someone else is going to hit it. It also puts your name on the finding rather than leaving it in a changelog entry. Tag me and I will gladly amplify it.

Thank you again. Genuinely good work. 🙏

MSKazemi:
**Merged** — this is in `main` as of 587ae83, and #174 is closed. 🎉

Your commit `01383c5` is in the history with your authorship, so it counts on your contributions graph as your work.

Closing this PR only for the mechanical reason in #170: a fork PR here cannot reach a mergeable state because CodeQL's default setup returns no results on one. **The change shipped; nothing was rejected.**

Thank you again — and the offer on #173 stands whenever you want it.

PR #186, opened 2026-09-08
playbook: add DnsResolutionFailure (#134 by @uuzzrm, closes #96)

### LINKED ISSUE #96: playbook: in-cluster DNS resolution failing

MSKazemi:
Adopts **#134 by @uuzzrm**, which was closed unmerged on 2026-08-15 in a batch with #135 and #137. Those two landed; **this one did not**, and nobody said so on the thread. That is a maintainer error, not a problem with the contribution. Their commit is carried here **unmodified, with their authorship intact**.

## The contribution

A `DnsResolutionFailure` playbook with event-message triggers for the log signatures DNS failures actually produce, plus three tests — two positive, and one proving it does **not** cross-fire with `NetworkPolicyBlocking` on a plain i/o timeout.

@uuzzrm deliberately left `detect:` null and [said why on #96](https://github.com/MSKazemi/kubeintellect/issues/96#issuecomment-3184913329): the failure shows up in application logs and CoreDNS SERVFAIL lines, neither of which reaches the kubectl watch sensorium — the pod stays `Running` with no Warning event. There was nothing honest to compile into a watch predicate. That is the correct call, it matches `NetworkPolicyBlocking`'s precedent, and given #114 (a playbook whose `detect:` block could never fire passed all nine gates) a contributor declining to invent a predicate is exactly the judgement this repo wants.

They also flagged that the upstream-resolver evidence lines were reasoned from the schema rather than observed, having no live cluster. Called out plainly, in the PR and the issue.

## Maintainer follow-up (second commit)

Only what moved underneath the branch while it sat:

- `test_the_inventory_is_actually_covered` hard-codes the counts so an empty registry cannot make the reachability tests pass vacuously. Added to `main` **after** #134 was written: 23 → 24 playbooks, 41 → 42 trigger regexes.
- `faq.md` playbook count and `AGENTS.md` server-suite count recollected by `check_doc_claims.py`.
- Two additive conflicts resolved by keeping both sides.

## Gates (Python 3.13)

| Gate | Result |
|---|---|
| `ruff check` | All checks passed |
| Server suite | 5486 passed, 23 skipped |
| playbook + reachability + RBAC | 109 passed |
| `check_doc_claims.py` | Doc claims now match the code |

Depends on #185 for a green required check.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -91,7 +91,7 @@ uv run ruff check packages/kubeintellect-server/app/ packages/ki-protocol/ packa
 # 2. Types — the workspace is at ZERO errors; keep it there.
 uv run mypy packages/kubeintellect-server/app packages/ki-protocol packages/kube-q/kube_q
 
-# 3. Server suite (5505 tests)
+# 3. Server suite (5510 tests)
 uv run python -m pytest tests/ -q
 
 # 4. kq CLI suite (749 tests)
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -29,7 +29,7 @@ The `v4/` platform is the current line and the only one you should deploy.
 - Conversational diagnosis correlating **kubectl + Prometheus + Loki**.
 - **Human-in-the-loop approval gate** with RBAC (superadmin / admin / operator / readonly) on
   every mutating action, enforced server-side at the mutating chokepoint.
-- **23 declarative failure playbooks** (`detect → investigate → remediate`), 20 of which compile
+- **24 declarative failure playbooks** (`detect → investigate → remediate`), 20 of which compile
   to zero-token detectors.
 - **Zero-token detection** — a detector engine over a kubectl `--watch` sensorium fires
   findings without invoking the LLM.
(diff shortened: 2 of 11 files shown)


MSKazemi:
Superseded by #187 — same two commits (@uuzzrm's unchanged, with their authorship), rebased onto `main` now that #184 and #185 have merged. Closing this one only to point at the rebased branch; the adoption is unchanged and is going in.

Issue #13, opened 2026-08-02
Add a new failure playbook


MSKazemi:
KubeIntellect ships declarative **failure playbooks** (YAML) that encode how to detect, investigate, and remediate a class of Kubernetes failure. Adding one is a drop-in file.

### Task
Add a playbook for a common failure not yet covered — see the uncovered list at the bottom — `ImagePullBackOff`, `Pending` PVC and HPA thrash are all already shipped — under `v4/packages/kubeintellect-server/app/agent/playbooks/`, following the existing schema, plus a test in `tests/`.

Read an existing playbook in that directory as a template. Perfect for learning the detect → investigate → fix flow. Comment here to claim it.

---

## Task metadata

| | |
|---|---|
| **Difficulty** | Beginner → Intermediate |
| **Estimated time** | 2–4 hours |
| **Skills** | YAML. Enough Kubernetes to describe one failure mode precisely. No Python needed — a playbook is a declarative `detect → investigate → remediate` file. |
| **Mentor** | @MSKazemi — ask on this issue, or in [Q&A](https://github.com/MSKazemi/kubeintellect/discussions/categories/q-a) |
| **Status** | 🟢 **Unclaimed** — comment "I'd like to take this" and it will be assigned to you |

Pick a failure mode nothing already covers. **Check against the tree rather than against this
issue** — a hard-coded list here goes stale the moment a playbook lands, which it has twice
already:

```bash
ls v4/packages/kubeintellect-server/app/agent/playbooks/*.yaml
grep -l "<your topic>" v4/packages/kubeintellect-server/app/agent/playbooks/*.yaml
```

**Still open** — verified by grepping every shipped playbook:

- **TLS / certificate expiry** — an expired secret behind an Ingress or webhook. Likely `detect: null`: an expired certificate surfaces in application logs and 5xx responses, not necessarily in a Warning Event.

**Already shipped since this list was written** — do not pick these:

- ~~PodDisruptionBudget blocking a drain or eviction~~ → [`pdb_blocking.yaml`](v4/packages/kubeintellect-server/app/agent/playbooks/pdb_blocking.yaml) ([#196](https://github.com/MSKazemi/kubeintellect/pull/196))
[...]

PR #196, opened 2026-09-09
feat(playbooks): add PDB-blocked eviction triage

biggdawg320:
## What & why

Related to #13; claimed the PDB topic in https://github.com/MSKazemi/kubeintellect/issues/13#issuecomment-5596045861. Leaves the umbrella issue open.

Add `PodDisruptionBudgetBlocking`, a triage guide for a voluntary eviction refused by a PDB. It requires a specific refusal plus current selector, budget and readiness evidence before recommending action. Zero allowed disruptions alone is not treated as a fault. Prefer restoring availability; policy changes and drain retries retain human approval.

The playbook deliberately uses `detect: null`: drain stderr is not a Kubernetes Warning Event, and Karpenter's PDB notices can be Normal events. Text routing recognizes those messages when supplied to the matcher, but the default Warning-only snapshot does not ingest them. This adds the guide and matching rules, not a new automatic drain detector or input channel. That limitation is documented.

Read-only PDB permissions are added to both shipped roles so the investigation can retrieve its required evidence. The RBAC coverage test maps the new resource. Generated documentation counts and inventory guards are updated.

## Type of change

- [x] New feature
- [x] Docs
- [x] Tests

## Scope

- v4 playbook, tests, chart read permissions, and generated documentation counts (including root AGENTS/CONTRIBUTING/ROADMAP).
- [x] Scoped to the current version.

## Checklist and validation

- [x] Positive and negative routing tests; four positive/registration cases failed before the playbook existed.
- [x] Triage-only detection explicitly tested as absent; no fabricated Event reason or zero-budget alert.
- [x] No cluster mutations added; human approval and secret handling preserved.
- [x] Ruff passes over the exact CONTRIBUTING/AGENTS scope.
- [x] Linux-targeted mypy passes: 193 source files. Native Windows mypy reports the existing `os.geteuid` reference.
- [x] Final focused tests pass on Python 3.12 and 3.13: **154 each**, including playbooks, routing, rendered Helm RBAC, documentation counts, and the rollback tests from newly merged upstream #195.
- [x] File modes (957 indexed files), syntax/encoding (565 Python files before upstream #195), roster and documentation-count gates passed. MkDocs built with existing missing-link warnings.
- [ ] Full suites green locally: **not claimed**. On Windows, both interpreters initially returned 82 server failures, 6 errors, 5339 passes and 106 skips. The unchanged base reproduced 80 failures and all 6 errors with identical names; the two additional failures were documentation counts, subsequently fixed and passing in the final focused runs. The CLI suite returned 10 failures / 739 passes; unchanged base reproduced all ten. Examples include Unix file-mode assumptions, shell execution and Windows SQLite file locking.

## Notes for reviewers

No live cluster was used. Helm 4.2.4 rendered the actual chart for the RBAC tests. Upstream #195 was merged and counts regenerated before final validation.

References: [Kubernetes eviction API](https://kubernetes.io/docs/concepts/scheduling-eviction/api-eviction/), [PDB configuration](https://kubernetes.io/docs/tasks/run-application/configure-pdb/), [Karpenter disruption messages](https://karpenter.sh/docs/concepts/disruption/).

Prepared and validated with OpenAI Codex under the account owner's direction; substantial AI assistance is disclosed.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -91,7 +91,7 @@ uv run ruff check packages/kubeintellect-server/app/ packages/ki-protocol/ packa
 # 2. Types — the workspace is at ZERO errors; keep it there.
 uv run mypy packages/kubeintellect-server/app packages/ki-protocol packages/kube-q/kube_q
 
-# 3. Server suite (5520 tests)
+# 3. Server suite (5533 tests)
 uv run python -m pytest tests/ -q
 
 # 4. kq CLI suite (749 tests)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -204,7 +204,7 @@ uv run mypy packages/kubeintellect-server/app packages/ki-protocol packages/kube
 
 # 3. Tests — two suites, run separately: the two `tests` packages collide
 #    under a single pytest invocation.
-uv run python -m pytest tests/ -q                              # server (~5520 tests)
+uv run python -m pytest tests/ -q                              # server (~5533 tests)
 cd packages/kube-q && uv run python -m pytest tests/ -q        # kq CLI (~749 tests)
 ```
 
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -29,7 +29,7 @@ The `v4/` platform is the current line and the only one you should deploy.
 - Conversational diagnosis correlating **kubectl + Prometheus + Loki**.
 - **Human-in-the-loop approval gate** with RBAC (superadmin / admin / operator / readonly) on
   every mutating action, enforced server-side at the mutating chokepoint.
-- **24 declarative failure playbooks** (`detect → investigate → remediate`), 20 of which compile
+- **25 declarative failure playbooks** (`detect → investigate → remediate`), 20 of which compile
   to zero-token detectors.
 - **Zero-token detection** — a detector engine over a kubectl `--watch` sensorium fires
   findings without invoking the LLM.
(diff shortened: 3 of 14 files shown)


MSKazemi:
**Merged.** Thank you, @biggdawg320 — this is an unusually strong first contribution, and I want to be specific about why, because most of what makes it good is judgement rather than code.

## What I verified

I did not take the description on trust. On your head SHA, Linux, Python 3.13:

```
ruff          All checks passed
mypy          no issues found in 193 source files
server suite  5519 passed, 23 skipped
kq suite      749 passed
doc claims    match the code
CI            19/19 checks green, mergeStateStatus CLEAN
```

**Your Windows report was accurate.** Both suites are fully green on Linux, so the 82/10 failures you saw were environmental exactly as you said. More importantly: **you ran the unchanged base as a control and reported the delta, and you did not tick "full suites green".** That is precisely the right way to report a result you cannot fully verify, and it is rarer than it should be.

I also ran three adversarial checks of my own:

```
1. PDB text matches only PodDisruptionBudgetBlocking   — no cross-fire out
2. no other playbook's trigger text routes to PDB       — no cross-fire in
3. Karpenter regex case-insensitive: PDB / pdb / PdB    — all match
```

## The two things that make this a good playbook

**`detect: null`, with a reason.** You wrote that an eviction refusal is an API response or drain stderr rather than a Warning Event, and that Karpenter reports PDB blockers as *Normal* events — neither of which the snapshot ingests. That is correct, and declining to invent a watch predicate is the single most valuable instinct here. [#114](https://github.com/MSKazemi/kubeintellect/pull/114) was a playbook whose `detect:` block **could never fire** and it passed all nine gates. A contributor who says "there is nothing honest to compile here" is doing the thing no gate can do.

**Zero allowed disruptions is not an incident.** Encoding that a PDB with `disruptionsAllowed: 0` is legitimate availability policy — and proving it with a negative test where a `kubectl get pdb` table showing `0` does *not* route to the playbook — is the difference between a useful playbook and one that cries wolf on every correctly-configured cluster. Your negative set is genuinely well chosen; the `Too Many Requests` case (a 429 with no PDB cause) is the one I would have asked for.

And the fix template refusing to delete the PDB or reach for `drain --disable-eviction` matters. A tool that suggests bypassing a disruption budget during an incident is worse than one that says nothing.

Two more things you did without being asked: read-only `policy/poddisruptionbudgets` on **both** shipped roles, and extending `_KINDS` in the RBAC-coverage test so the derived list picks it up. That test exists because a playbook once shipped without the permissions it needed and nothing noticed.

## Your PR also settled an open question about this repo

**[#170](https://github.com/MSKazemi/kubeintellect/issues/170)** claimed fork PRs could never reach a mergeable state, because the branch required code-scanning results that a fork PR cannot produce. Recent outside contributions have been landed by carrying the commit onto an integration branch as a workaround.

Yours is a genuine cross-repo fork PR (`isCrossRepository: true`), it reached **19/19 green and `CLEAN`**, and it merged normally. So the blocker is gone, and #170 closes on your evidence. Your commits are in `main` under your own authorship, and this is a real merge rather than the workaround.

## Credit

You are in [`CHANGELOG.md`](https://github.com/MSKazemi/kubeintellect/blob/main/CHANGELOG.md), the [README contributors table](https://github.com/MSKazemi/kubeintellect#contributors) and `.all-contributorsrc` (#197), and you will be named in the next release notes. The changelog entry records your two judgement calls rather than the feature, because those are what the next playbook author should copy.

The Codex disclosure is exactly what we ask for and counts for you, not against you — it tells a reviewer where to look, nothing more.

---

If you would like more: **[#13](https://github.com/MSKazemi/kubeintellect/issues/13) stays open** and two of the three genuinely uncovered failure modes are still unclaimed — **TLS/certificate expiry** and a **StatefulSet-specific stuck rollout** (ordinal blocked behind a pending PVC or a failing readiness gate). You now know this schema and its traps better than almost anyone; either is yours if you want it.

And one entirely optional thing, with no bearing on the review since it is already merged: PDB-blocked drains are a genuinely under-documented failure, and the "nothing looks broken, the drain just never finishes" shape is a good short write-up. If you feel like posting about it, I would be glad to amplify it — it puts your name on the finding rather than leaving it in a changelog line. A ⭐ or a share is welcome too if the project turns out useful to you. Entirely your call.

Thank you again — genuinely good work. 🙏

PR #198, opened 2026-09-09
docs(triage): explain fork workflow approval blockers

biggdawg320:
## What & why

A first-time fork PR can show green greeting/label jobs while its real workflows await maintainer approval. TRIAGE.md now explains how to recognize that state, review and approve the runs, verify required checks on the current revision, and investigate other merge blockers without assuming a CodeQL failure.

Completes the small documentation follow-up described in #170 and linked from #15. Both issues are already closed; this does not reopen their original scope. Uses #196 as the verified example.

## Type of change

- [x] 📖 Docs

## Scope

- Root TRIAGE.md only: 37 added lines. No application, workflow or protection-setting changes.
- [x] Scoped to the maintainer-requested documentation follow-up.

## Validation

- `python scripts/check_doc_claims.py` from v4: passed.
- `mkdocs build` from v4: passed with existing missing-page/link warnings. TRIAGE.md is outside that site's source, so its Markdown rendering and local reference were checked separately.
- Rendered TRIAGE.md with the installed Markdown parser; checked the new heading and local required-check reference.
- `git diff --check origin/main <published-commit>`: passed.
- Checked workflow-approval instructions against current GitHub documentation and the maintainer's measured #170/#196 results.
- Application suites, lint and type checks were not rerun for this documentation-only change; no claim of fresh full-suite validation.

## Notes for reviewers

Substantial Codex assistance was used to research, draft and verify this change. The distinction between workflow approval and code-review/merge approval is explicit. No new CLI flags or simulated command output are introduced.

PR #202, opened 2026-09-09
feat(playbooks): add StatefulSet rollout triage

biggdawg320:
## What & why

Adds StatefulSetRolloutStuck as a triage-only guide for a blocked ordinal, following the scope agreed in #13 (comment 5597827582). Controller Pod/Claim creation failures now offer it as a candidate. The guide requires verified ownership and timestamped progress evidence, and checks PVC/readiness causes, OnDelete, intentional partitions and minReadySeconds before diagnosing a stall.

Related to #13; leaves the umbrella issue and TLS topic open. Claim: https://github.com/MSKazemi/kubeintellect/issues/13#issuecomment-5598000194

## Type of change

- [x] ✨ New playbook
- [x] 📖 Docs
- [x] 🧪 Tests

## Scope

- v4 playbook, 14 focused tests, manual inventory assertions (26 playbooks / 46 trigger regexes), and generated count updates. Root AGENTS/CONTRIBUTING/ROADMAP changes are generated counts only.
- Based on main 77816be, including #200 and #201.
- No new permissions, telemetry collection or automatic stall detector. Existing rendered chart permissions cover the read commands.

## Checklist and validation

- [x] Before adding YAML: 3 failed / 11 passed in the new test file. After: all 14 pass.
- [x] Python 3.12 and 3.13: 165 focused tests pass each (routing, playbooks, doc claims, rendered Helm RBAC).
- [x] Exact CI lint scope passes; Linux-targeted mypy passes, 193 files.
- [x] Doc claims match; MkDocs builds with existing missing-link/page warnings.
- [x] File-mode gate: 961 files; syntax/encoding gates: 568 Python files; roster: 11 contributors; published diff whitespace check passes. File-mode script used an LF-normalized copy under Git Bash.
- [ ] Full application suites green locally — **not claimed**. Both Python versions on Windows: server **80 failed / 5384 passed / 92 skipped / 6 errors**; CLI **10 failed / 739 passed**. Every failure/error node ID matches our previously recorded unchanged Windows baseline, with no new failure names. This is comparison against that prior baseline run, not a new baseline run on this revision. Upstream Linux CI remains the full-suite check.
- [x] Human approval, dry-run/diff, quorum and persistent-data safeguards retained. No cluster or LLM calls used in validation.

## Notes for reviewers

`detect: null` is intentional. A snapshot cannot establish elapsed rollout progress or correlate ownership through WatchPredicate. Generic Pending/Unhealthy and normal rollout-status output do not match this guide. PVC/readiness stalls without a creation error remain investigation cases; this PR does not add automatic routing for them. Multiple candidate playbooks may co-fire, per #201.

Tests cover real controller creation-message forms, success, ordinary waiting, unrelated ReplicaSet/readiness/scheduling errors, empty input and a multi-line negative to prevent combining a success with another object's failure. Message templates were checked against Kubernetes stateful_pod_control.go. Recovery guidance explicitly notes that reverting a bad template alone may not recover a StatefulSet, while preserving approval before any affected-pod recreation.

Substantial Codex assistance was used to research, implement and run validation.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -91,7 +91,7 @@ uv run ruff check packages/kubeintellect-server/app/ packages/ki-protocol/ packa
 # 2. Types — the workspace is at ZERO errors; keep it there.
 uv run mypy packages/kubeintellect-server/app packages/ki-protocol packages/kube-q/kube_q
 
-# 3. Server suite (5545 tests)
+# 3. Server suite (5562 tests)
 uv run python -m pytest tests/ -q
 
 # 4. kq CLI suite (749 tests)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -204,7 +204,7 @@ uv run mypy packages/kubeintellect-server/app packages/ki-protocol packages/kube
 
 # 3. Tests — two suites, run separately: the two `tests` packages collide
 #    under a single pytest invocation.
-uv run python -m pytest tests/ -q                              # server (~5545 tests)
+uv run python -m pytest tests/ -q                              # server (~5562 tests)
 cd packages/kube-q && uv run python -m pytest tests/ -q        # kq CLI (~749 tests)
 ```
 
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -29,7 +29,7 @@ The `v4/` platform is the current line and the only one you should deploy.
 - Conversational diagnosis correlating **kubectl + Prometheus + Loki**.
 - **Human-in-the-loop approval gate** with RBAC (superadmin / admin / operator / readonly) on
   every mutating action, enforced server-side at the mutating chokepoint.
-- **25 declarative failure playbooks** (`detect → investigate → remediate`), 20 of which compile
+- **26 declarative failure playbooks** (`detect → investigate → remediate`), 20 of which compile
   to zero-token detectors.
 - **Zero-token detection** — a detector engine over a kubectl `--watch` sensorium fires
   findings without invoking the LLM.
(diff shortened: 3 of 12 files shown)


MSKazemi:
**Merged.** Third contribution today, and the strongest of the three.

## What I verified

Routing, on your head SHA — every case, including the multi-line contamination one you called out:

```
OK  want=True  :: 'create Pod web-0 in StatefulSet web failed error: pods "web-0" is forbidden'
OK  want=True  :: 'create Claim data-web-0 for Pod web-0 in StatefulSet web failed error: …'
OK  want=False :: 'create Pod web-0 in StatefulSet web successful'
OK  want=False :: 'Waiting for pod default/web-0 to be Running and Ready'
OK  want=False :: 'Readiness probe failed: connection refused'
OK  want=False :: 'create Pod api-1 in ReplicaSet api failed error: quota exceeded'
routing errors: 0
multi-line success + unrelated-failure fires SS?  False
```

That last one is the case most people would miss. A regex scanning a whole events stream can happily pair *"create Pod … successful"* from one object with *"failed error:"* from another; anchoring on the full controller message form prevents it, and you tested for it rather than hoping.

Gates: `ruff` clean, `mypy` 193 files, doc claims match, **5539 passed / 23 skipped**, CI 17/17 `CLEAN`.

## Three things that make this better than the last one

**You ran the red-green yourself and reported both numbers.** *"Before adding YAML: 3 failed / 11 passed. After: all 14 pass."* That is the check this repo asks for and the one contributors most often skip, and you did it unprompted.

**You checked the message forms against `stateful_pod_control.go`** rather than inventing plausible text. The difference between a trigger that matches what Kubernetes actually emits and one that matches what it *seems like* it should emit is invisible until an incident, which is exactly the #114 failure mode.

**The domain detail is genuinely expert.** Three things in the investigation steps that most Kubernetes engineers get wrong:

- *"Do not infer ownership from a name prefix or assume ordinals start at zero"* — `ordinals.start` exists and almost nobody accounts for it.
- *"Ordered creation waits on lower ordinals, whereas rolling updates normally advance from higher ordinals"* — correct, and the opposite of most people's intuition.
- *"A Running phase does not imply Ready; distinguish a failing container probe from an unmet custom readiness gate"* — readiness gates are the quiet cause of exactly this stall.

And deferring to `PvcPending`, `ReadinessProbeFailing`, `ImagePullBackOff` and `QuotaExceeded` for the underlying cause — while saying multiple candidates are expected — is precisely the model from #201. That doc landed roughly an hour before you used it, which is about the best feedback a doc change can get.

## On your validation note

> This is comparison against that prior baseline run, not a new baseline run on this revision.

That distinction is genuinely precise, and most people would not have drawn it. For the record, both suites are green on Linux — **5539 server, 749 kq** — so your Windows failures are environmental as you concluded.

## Credit

Your README entry and `CHANGELOG` will pick this up; I will fold it in shortly, as with #198.

---

**[#13](https://github.com/MSKazemi/kubeintellect/issues/13) stays open with TLS / certificate expiry unclaimed**, exactly as you left it. Worth noting the same reasoning probably lands there too: an expired certificate behind an Ingress surfaces in application logs and 5xx responses rather than a Warning Event, so `detect: null` is likely honest there as well — but that is yours to assess, or someone else's to take.

Thank you. Three merges and a design discussion in one morning is a lot, and the quality did not drop across any of them. 🙏

PR #207, opened 2026-09-10
fix(coordinator): ground command examples and sequence name discovery

biggdawg320:
## What & why
The coordinator's parallel example suggested describing a node before its name was known, and its command examples taught angle-bracket placeholders. This addresses the standalone **prompt-side** contribution offered in #173: discovery must finish before dependent calls, identifiers must come from evidence, and examples use concrete illustrative values with explicit substitution guidance. The truncation warning's shared clause is updated consistently.

Related to #173; **does not close it**. The pre-execution guard and a model/tool-call sequencing regression remain separate work. Prompt instructions alone cannot guarantee model compliance.

## Type and scope
- [x] Bug fix / prompt guidance, v4 only
- [x] Existing approval, RBAC and kubectl safety gates preserved
- [x] No secret values introduced

## Validation
- Coordinator/routing, parallel isolation and kubectl tests: 105 passed (Python 3.12).
- Shared truncation/prompt tests: 35 passed on Python 3.12 and 3.13 after correcting a shared-clause mismatch caught by the full run.
- Ruff passes; Linux-targeted mypy passes (193 files); syntax (Python 3.13), encoding and contributor roster gates pass; doc claims match; git diff --check passes.
- Review correction: the original local audit missed `<node>` and `<N>` and its zero result was too narrow. A committed regression test now detects command metavariables and standalone flag values; it failed on all four omitted examples before correction and passes afterward. Cordon/drain/uncordon now use a verified-name example; grace period is expressed in prose. 136 focused tests pass on Python 3.12 and 3.13, including the new gate. Generated test counts refreshed to5563. This checks prompt content, not live LLM behavior.
- Full server suites ran on both interpreters before the shared-clause correction: each 81 failed, 5369 passed, 106 skipped, 6 errors. One failure was the clause mismatch fixed above; all other failure/error names match the previously recorded Windows baseline (80 failures/6 errors). No fresh full baseline rerun or claim of a green full suite.
- Full CLI suites: each 739 passed, 10 failed; failure names match the prior Windows baseline.
- No live cluster or paid LLM evaluation performed. The added test is non-executable (git mode100644) and has no shebang; the full file-mode gate was not rerun for this revision.

## Notes for reviewers
Substantial AI assistance: Codex performed implementation and local validation under Jim Hall's direction. This is intentionally the independently offered prompt contribution; it does not implement the runtime guard or guarantee that placeholders can never reach a tool.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -91,7 +91,7 @@ uv run ruff check packages/kubeintellect-server/app/ packages/ki-protocol/ packa
 # 2. Types — the workspace is at ZERO errors; keep it there.
 uv run mypy packages/kubeintellect-server/app packages/ki-protocol packages/kube-q/kube_q
 
-# 3. Server suite (5562 tests)
+# 3. Server suite (5563 tests)
 uv run python -m pytest tests/ -q
 
 # 4. kq CLI suite (749 tests)
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -91,7 +91,7 @@ uv run ruff check packages/kubeintellect-server/app/ packages/ki-protocol/ packa
 # 2. Types — the workspace is at ZERO errors; keep it there.
 uv run mypy packages/kubeintellect-server/app packages/ki-protocol packages/kube-q/kube_q
 
-# 3. Server suite (5584 tests)
+# 3. Server suite (5590 tests)
 uv run python -m pytest tests/ -q
 
 # 4. kq CLI suite (749 tests)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,30 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ### Fixed
 
+- **The coordinator prompt taught the model to emit `<ns>` literally**
+  (`app/agent/nodes/coordinator.py`, `app/tools/output_policy.py`, fixed by
+  [@biggdawg320](https://github.com/biggdawg320), #207, closes the prompt half of #173).
+  Every kubectl example in the system prompt used metavariables — `-n <ns>`, `<name>`,
+  `--grace-period=<N>` — eighteen times over. A model copies the shape it is shown, so it
+  emitted `<ns>` verbatim, the kubectl guard rejected the command for containing shell
+  metacharacters, and the run failed. The examples now use concrete illustrative values
+  (`shop`, `payments-api`, `worker-1`) with an explicit instruction that they are
+  illustrative, that they must be substituted from discovery, and that a name must never
+  be guessed to make a command executable.
+
+  `describe node` was also dropped from the "Parallel (always)" batch, which had told the
+  model to describe a node in the same parallel batch that discovers which node to look at.
+
+  `v4/tests/test_coordinator_command_examples.py` keeps the examples grounded. It counts a
+  line as a command example if it names a CLI this project drives or carries a flag or a
+  `key=` directive — not only if it contains the word `kubectl`, which would have walked
+  past two of the lines this change repaired — and it runs over the Cortex gather and
+  synthesis prompts too, since the truncation clause is spliced into all three.
+
+  Still open in #173: `app/tools/kubectl_errors.py` hands metavariables back to the model
+  on the error path (`re-run \`kubectl get pods -n <ns>\``), with hints enabled by default.
+  That is the same failure at the moment the model is most likely to copy a suggestion.
+
 - **`kubeintellect service start` and `service stop` exited 0 after systemd refused**
   (`app/cli.py`, reported and fixed by [@Ryota-Di](https://github.com/Ryota-Di), #208).
   Both called `subprocess.run(...)` and discarded the result, so a unit that failed to
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -204,7 +204,7 @@ uv run mypy packages/kubeintellect-server/app packages/ki-protocol packages/kube
 
 # 3. Tests — two suites, run separately: the two `tests` packages collide
 #    under a single pytest invocation.
-uv run python -m pytest tests/ -q                              # server (~5584 tests)
+uv run python -m pytest tests/ -q                              # server (~5590 tests)
 cd packages/kube-q && uv run python -m pytest tests/ -q        # kq CLI (~749 tests)
 ```
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -204,7 +204,7 @@ uv run mypy packages/kubeintellect-server/app packages/ki-protocol packages/kube
 
 # 3. Tests — two suites, run separately: the two `tests` packages collide
 #    under a single pytest invocation.
-uv run python -m pytest tests/ -q                              # server (~5562 tests)
+uv run python -m pytest tests/ -q                              # server (~5563 tests)
 cd packages/kube-q && uv run python -m pytest tests/ -q        # kq CLI (~749 tests)
 ```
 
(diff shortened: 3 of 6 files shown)


MSKazemi:
@biggdawg320 — your fourth, and it is green on all nine required checks. Reviewed properly; two things you got right that I want to name, and one real finding.

## The diagnosis is better than the issue's

#173 described the symptom — the coordinator emits `<ns>` and the safety gate refuses it. You identified the *cause*: the prompt was **teaching** the placeholder by showing it in twelve worked examples. A model copies the shape of the examples it is given. Nothing in the issue said that, and it is the difference between a patch and a fix.

**And you found a second bug nobody had filed.** The parallel example was:

```
Parallel (always):   (get pods) + (get events) + (describe node)
```

You cannot describe a node in the same batch that discovers it — the name does not exist yet. That example instructed the model to do the exact thing the "Sequential (only)" line beneath it forbids, and it had been sitting there contradicting itself. Removing `describe node` from the parallel batch is correct.

**The line I would have asked you to add, you already added:**

> do not guess names to make a command executable. If discovery does not identify the target, report the missing evidence.

That is the whole risk of this change in one sentence. Replacing `<ns>` with `shop` removes a placeholder the safety gate *catches* — and the obvious way for that to go wrong is the model substituting a plausible invented name instead, which the gate does **not** catch, because `-n shop` is a perfectly well-formed command that quietly returns nothing on a cluster with no `shop` namespace. That would trade a loud failure for a silent wrong answer, which is safety invariant #1 exactly. You saw it and wrote the counter-instruction, plus the explicit "they are not evidence that those resources exist" note.

## The finding: three examples still carry a placeholder, and they are the mutating ones

The Node Drain block was missed. On your branch (`90d5a9f`), inside `_COORDINATOR_SYSTEM`:

```console
$ grep -n 'kubectl \(cordon\|drain\|uncordon\)' app/agent/nodes/coordinator.py
482:       kubectl cordon <node>
484:       kubectl drain <node> --ignore-daemonsets --delete-emptydir-data
490:       kubectl uncordon <node>
```

Plus `--grace-period=<N>` on line 485. These are the same shape #173 is about, and they are the **highest-stakes** commands in the prompt — cordon and drain mutate the cluster, so they are the ones that reach the HITL gate. An operator approving a drain plan is the worst place for `<node>` to survive.

Not a blocker; the PR is a clear net improvement and I would rather have it in than hold it for this. But it is the difference between "most examples fixed" and "the class is closed".

## The thing that would make this permanent

You ran the audit **locally** — *"12 placeholder-bearing examples before and zero after"*. That number is the most valuable artifact in this PR and it currently lives only in the description. Encoded as a test, it becomes a gate:

```python
def test_no_command_example_teaches_an_unresolved_placeholder():
    hits = re.findall(r"kubectl[^\n]*<[a-z][a-z0-9_.-]*>[^\n]*", _COORDINATOR_SYSTEM)
    assert hits == [], hits
```

That is red on `main`, red on your branch (it catches the three drain lines), and green once they are fixed. It is also the only form of this work that survives the next person adding an example. If you want to take that as a follow-up — the drain block plus the guard — it would close #173's prompt half properly.

## On how you reported it

Worth saying plainly, because it is rarer than it should be: you wrote *"Related to #173; **does not close it**"*, *"Prompt instructions alone cannot guarantee model compliance"*, and *"No fresh full baseline rerun or claim of a green full suite"* — then ran the unchanged base as a control to separate your changes from the pre-existing Windows failures. That is a contributor telling me the limits of their own evidence before I ask. It makes your PRs cheap to review and it is why this one got read carefully rather than skimmed.

The `Tests (server)` gate catching your shared-clause mismatch is also the system working — `test_a_marker_nobody_is_looking_for_is_not_a_warning.py:158` asserts `TRUNCATION_CLAUSE in _COORDINATOR_SYSTEM`, so the two copies cannot drift. You hit it and fixed it rather than working around it.

---

Merge decision is with the maintainer and I expect it to be yes. Your [wall row](https://github.com/MSKazemi/kubeintellect#contributors) gets rewritten to include this once it lands.

And the standing optional one: if you ever write up *"an AI ops tool was teaching its own model to emit invalid commands, because the examples in its prompt were templates"* — that generalises to essentially every prompt-engineered agent, and it is a better post than it is a changelog line. Tag me and I will amplify it. Only if you want to. 🙏

biggdawg320:
Thanks for catching the drain block. Fixed in 071e86c: all three node examples now use an explicitly illustrative name that must be replaced with the same verified target, and grace-period guidance uses prose rather than `<N>`. The committed test catches both command placeholders and standalone flag values; I ran it red on all four omissions, then green after correction. 136 focused tests pass on both Python3.12 and3.13; Ruff, syntax, encoding and generated doc counts pass.

My original local audit used too narrow a placeholder vocabulary, so its zero result was misleading; I corrected the PR description. Also, to keep the evidence accurate, I compared full-suite failure names with the previously recorded Windows baseline, not a freshly rerun unchanged base. The runtime guard remains separate scope.

MSKazemi:
Merged — thank you. This one goes deeper than the issue it closes.

Your change is on `main` here: https://github.com/MSKazemi/kubeintellect/blob/3f13e67915e98a97cf6ceb2c770ce7ba4c0e6dad/v4/packages/kubeintellect-server/app/agent/nodes/coordinator.py#L344-L346

#173 reads like a validation bug — the guard rejected `<ns>`, so tighten something around the guard. Your diagnosis is the better one: the prompt was *teaching* the placeholder shape, eighteen times, and the model was doing exactly what it had been shown. Demonstrating the shape you then reject is the kind of defect that survives review indefinitely because every individual line looks fine. Removing `describe node` from the "Parallel (always)" batch is the same quality of catch — it told the model to describe a node in the same parallel batch that discovers which node to describe.

The part that makes this stick is the gate. Grounding the examples fixes today's prompt; `test_coordinator_command_examples.py` is what stops the next person reintroducing it, and that is the difference between a fix and a repair.

Verified before merging, on the merged tree: ruff clean, mypy clean across 193 files, server suite 5566 passed, the marker-parity suite 35 passed, and the file-mode, syntax, encoding, roster and doc-claims gates green. I confirmed your gate genuinely fails on the unpatched prompt first — 18 hits — rather than assuming it.

Two changes of mine on top, so you know what moved under your name:

1. I widened your gate's regex. It anchors on the literal word `kubectl` appearing earlier in the line, which walks straight past two lines your own PR repaired: the `TARGETED: namespace=...` directive, and the `-n <namespace>` inside the truncation clause. I checked rather than assumed — putting the old shape back and running your pattern returned *no hits at all*, so a regression on either line would have shipped green. A line now counts as a command example if it names a CLI, carries a flag, or has a `key=` directive, and it runs over the Cortex gather and synthesis prompts too, since the truncation clause is spliced into all three. `<none>`, `<findings>`, `<state>` and `<metric>` are excluded by name — `<none>` is what kubectl prints in an empty ENDPOINTS column, and a gate that flagged it would push someone to reword a correct instruction to appease it.

2. I shortened the truncation warning to "narrow the query with `-n`, `-l`, or `--tail`". I want to be clear that I kept your change rather than reverting it: my first instinct was that the clause was out of scope, and checking showed the opposite — reverting it puts `-n <namespace>` straight back into the prompt, which is the shape this PR exists to remove. You were right to change it. I only made it shorter, because that line is read by an operator mid-incident.

One thing your PR correctly says it does not close, now recorded in the changelog so #173 does not look finished: `app/tools/kubectl_errors.py` still hands metavariables back to the model on the error path — `re-run \`kubectl get pods -n <ns>\`` — with `KUBECTL_ERROR_HINTS_ENABLED` defaulting to true. That is the same failure arriving at the moment the model is most likely to copy a suggestion verbatim. If you want it, it is yours; extending your gate to cover `_PATTERNS` would close the class properly.

Your #210 is in review separately — I owe you a proper answer there and it is coming.

Entirely optional, and the merge and the credit stand either way: if the project is useful to you, a star or a mention helps other people find it.

PR #208, opened 2026-09-10
fix: propagate service start and stop failures

### LINKED ISSUE #189: `kubeintellect service start` exits 0 when systemd refuses to start the unit
**`kubeintellect service start` prints a failure and then exits `0`.** Reproduced on 2026-09-09 against `main`:

```console
$ kubeintellect service start
Failed to start kubeintellect.service: Unit kubeintellect.service not found.
$ echo $?
0

$ kubeintellect service stop
Failed to stop kubeintellect.service: Unit kubeintellect.service not loaded.
$ echo $?
0
```

`systemctl` itself exits **5** here. We throw that away.

## Why it matters

A zero exit code is a claim that the thing you asked for happened. Anything that scripts this command — a provisioning playbook, a CI job, a `&&` chain, a wrapper unit — reads `0` and moves on to talk to a server that is not running. That is the same shape as the project's first safety invariant (*never report a result you do not have*), just at the CLI layer instead of the diagnosis layer.

It is also **inconsistent with its own neighbours**. `install` and `uninstall`, ten lines above in the same function, do this correctly.

## Where

[`v4/packages/kubeintellect-server/app/cli.py`](https://github.com/MSKazemi/kubeintellect/blob/main/v4/packages/kubeintellect-server/app/cli.py) — `cmd_service()`, around lines 782–789:

```python
elif action == "start":
    subprocess.run(["systemctl", "--user", "start", _SERVICE_NAME])
elif action == "stop":
    subprocess.run(["systemctl", "--user", "stop", _SERVICE_NAME])
elif action == "status":
    subprocess.run(["systemctl", "--user", "status", _SERVICE_NAME])
elif action == "logs":
    subprocess.run(["journalctl", "--user", "-u", _SERVICE_NAME, "-f", "--no-pager"])
```

Compare with `install` in the same function, which is the pattern to follow:

```python
result = _install_service()
if not result.ok:
    _print_service_failure(result)
    sys.exit(1)
```

## What to do

Propagate the child's exit code for `start` and `stop`. Capture the return code and `sys.exit()` with it (or with `1`) when it is non-zero.

**Please think about `status` and `logs` separately rather than treating all four the same** — that judgement is most of the value of this issue:

- `status` — `systemctl status` exits **3** for a stopped-but-installed unit. That is a *successful query with a negative answer*, not a failure to query. Making `kubeintellect service status` exit 3 may be right (it mirrors systemctl) or wrong (it breaks `if kubeintellect service status; then`). Say which you chose and why.
- `logs` — `journalctl -f` is interrupted by Ctrl-C, which is a normal exit for a follow command, not an error.

A PR that fixes `start`/`stop`, states a reasoned position on `status`/`logs`, and does nothing else is exactly right. **"I fixed start/stop and here is why I left status alone" is a complete answer.**

## Acceptance criteria

- [ ] `kubeintellect service start` exits non-zero when `systemctl` does
- [ ] Same for `stop`
- [ ] A test that **fails before your change and passes after** — patch `subprocess.run` to return a non-zero `returncode` and assert the command exits non-zero. Run it against unpatched code first and watch it fail; if it passes before your change, it is not testing anything.
- [ ] A short note in the PR on what you decided for `status` and `logs`
- [ ] `v4/tests/` suite still green

## Getting started

```bash
git clone https://github.com/MSKazemi/kubeintellect && cd kubeintellect/v4
uv sync
uv run python -m pytest tests/ -q          # ~5500 tests, no cluster or API key needed
```

You do **not** need a Kubernetes cluster, an API key, or systemd to work on this — the test patches `subprocess.run`. Existing CLI tests in `v4/tests/` show the pattern.

## Notes

Found while re-measuring the `ruff` 0.16 backlog (#75) — `PLW1510` flags all four of these as `subprocess.run` without an explicit `check=`. Most of that family is genuinely cosmetic; **these four are not**, which is a decent argument for reading a lint backlog rather than batch-fixing it.

## Task metadata

| | |
|---|---|
| **Difficulty** | Beginner |
| **Estimated time** | ~1 hour |
| **Skills** | Python. No Kubernetes, no API key, no systemd required. |
| **Mentor** | @MSKazemi — ask right here, or in [Q&A](https://github.com/MSKazemi/kubeintellect/discussions/categories/q-a) |
| **Status** | 🟢 **Unclaimed** — comment "I'd like to take this" and it is yours |

Contributors are credited in `CHANGELOG.md`, the README contributors table and the release notes.

Ryota-Di:
## What & why

`kubeintellect service start` and `kubeintellect service stop` now propagate a non-zero exit code when the underlying `systemctl` command fails.

Previously, `subprocess.run(...)` was called without inspecting its result, so `systemctl` could fail while the kubeintellect CLI still exited successfully with code 0.

This PR:

* checks the `systemctl` return code for `service start` and `service stop`
* propagates the underlying non-zero exit code
* adds regression coverage for both commands
* removes the repaired `start` / `stop` calls from `_REVIEWED`
* updates the remaining `_REVIEWED` reasoning for `status` and `logs` to explicitly address exit-code semantics
* updates the documented server test count after adding the two parametrized regression cases

### Why `status` and `logs` are unchanged

I left `service status` and `service logs` unchanged intentionally.

For `status`, a non-zero `systemctl status` code can describe a valid queried state such as an inactive service, so it should not be blindly treated as failure of the status query itself.

For `logs`, `journalctl -f` is an interactive follow command whose normal lifecycle includes user interruption, so its eventual subprocess return code does not directly represent whether log viewing was successfully started.

Their `_REVIEWED` entries now document these exit-code considerations rather than only noting that stdout/stderr are visible to the user.

### Testing

The regression test was first run against the unmodified implementation and failed as expected for both `start` and `stop`:

```text
FAILED ...[start] - Failed: DID NOT RAISE SystemExit
FAILED ...[stop] - Failed: DID NOT RAISE SystemExit
2 failed
```

After the fix:

```text
17 passed
```

Additional validation:

```text
test_doc_claims_match_code: 1 passed
ruff: All checks passed!
mypy: Success: no issues found in 193 source files
```

I also ran the full server test suite locally. The remaining failures were unrelated to this change and specific to the local macOS environment: the demo GIF tests require a supported monospace font such as DejaVu Sans Mono, and three file-mode tests hit BSD `chmod` behavior.

Closes #189.

## Type of change

* [x] 🐛 Bug fix
* [ ] ✨ New feature
* [x] 📖 Docs
* [ ] 🧹 Refactor / chore
* [ ] ⚡ Performance
* [x] 🧪 Tests

## Scope

* Version directory touched: `v4/` (plus root-level test-count docs)

## Checklist

* [ ] New behavior has both a **happy-path** and an **error-path** test
* [ ] **Adding a playbook?** `triggers:` and `detect:` are two independent features — a test proves
  the compiled `detect:` predicate **fires** on a realistic event and does **not** fire on a
  neighbouring one. The counts and the schema check prove neither. (`kind:` is the observation
  channel — `Pod` / `Event` / `Node` — never the Kubernetes object; use `involved_kind:` to
  narrow the subject.)
* [ ] **Every mutating/write operation keeps its dry-run + diff + human-approval (HITL) gate** (safety invariant)
* [ ] Secret values are never logged or returned (key names only)
* [ ] `uv run pytest` passes locally
* [x] `uv run ruff check .` passes locally
* [x] `uv run mypy src` passes locally
* [x] Docs updated if behavior/CLI/flags changed

## Notes for reviewers

The full server suite was run locally on macOS. The failures described above appear environment-specific; the tests directly covering this change, doc-claim validation, lint, and type checking pass locally.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -162,7 +162,9 @@
       "avatar_url": "https://avatars.githubusercontent.com/u/53958043?v=4",
       "profile": "https://github.com/Ryota-Di",
       "contributions": [
-        "inflight"
+        "bug",
+        "code",
+        "test"
       ]
     },
     {
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -91,7 +91,7 @@ uv run ruff check packages/kubeintellect-server/app/ packages/ki-protocol/ packa
 # 2. Types — the workspace is at ZERO errors; keep it there.
 uv run mypy packages/kubeintellect-server/app packages/ki-protocol packages/kube-q/kube_q
 
-# 3. Server suite (5562 tests)
+# 3. Server suite (5564 tests)
 uv run python -m pytest tests/ -q
 
 # 4. kq CLI suite (749 tests)
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -91,7 +91,7 @@ uv run ruff check packages/kubeintellect-server/app/ packages/ki-protocol/ packa
 # 2. Types — the workspace is at ZERO errors; keep it there.
 uv run mypy packages/kubeintellect-server/app packages/ki-protocol packages/kube-q/kube_q
 
-# 3. Server suite (5578 tests)
+# 3. Server suite (5584 tests)
 uv run python -m pytest tests/ -q
 
 # 4. kq CLI suite (749 tests)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,26 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Fixed
+
+- **`kubeintellect service start` and `service stop` exited 0 after systemd refused**
+  (`app/cli.py`, reported and fixed by [@Ryota-Di](https://github.com/Ryota-Di), #208).
+  Both called `subprocess.run(...)` and discarded the result, so a unit that failed to
+  start reported success to the caller — and to any script or CI step that trusted the
+  exit code. They now propagate systemd's own return code.
+
+  The fix is narrower than it looks, and deliberately so: `service status` and
+  `service logs` keep discarding theirs, because `systemctl status` returns non-zero for
+  a perfectly valid inactive unit and `journalctl -f` returns non-zero when the operator
+  presses Ctrl-C. Treating either as a failure would have traded one wrong answer for
+  another. Those two entries stay in the audit allowlist with that reasoning written out.
+
+  `systemctl --user start` and `stop` were added to `_MUST_STAY_CHECKED`, and the success
+  path gained a test of its own: the failure test pins the code that is propagated, but
+  nothing held the other side, so `sys.exit(proc.returncode or 1)` — the shape a later
+  cleanup reaches for — would have turned every successful start into exit 1 with the
+  suite still green.
+
 ## [2.5.0] – 2026-09-12
 
 ### Added
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -204,7 +204,7 @@ uv run mypy packages/kubeintellect-server/app packages/ki-protocol packages/kube
 
 # 3. Tests — two suites, run separately: the two `tests` packages collide
 #    under a single pytest invocation.
-uv run python -m pytest tests/ -q                              # server (~5562 tests)
+uv run python -m pytest tests/ -q                              # server (~5564 tests)
 cd packages/kube-q && uv run python -m pytest tests/ -q        # kq CLI (~749 tests)
 ```
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -204,7 +204,7 @@ uv run mypy packages/kubeintellect-server/app packages/ki-protocol packages/kube
 
 # 3. Tests — two suites, run separately: the two `tests` packages collide
 #    under a single pytest invocation.
-uv run python -m pytest tests/ -q                              # server (~5578 tests)
+uv run python -m pytest tests/ -q                              # server (~5584 tests)
 cd packages/kube-q && uv run python -m pytest tests/ -q        # kq CLI (~749 tests)
 ```
 
--- a/v4/packages/kubeintellect-server/app/cli.py
+++ b/v4/packages/kubeintellect-server/app/cli.py
@@ -780,9 +780,13 @@ def cmd_service(args: argparse.Namespace) -> None:
             sys.exit(1)
         print(_ok("✓  Service removed."))
     elif action == "start":
-        subprocess.run(["systemctl", "--user", "start", _SERVICE_NAME])
+        proc = subprocess.run(["systemctl", "--user", "start", _SERVICE_NAME])
+        if proc.returncode != 0:
+            sys.exit(proc.returncode)
     elif action == "stop":
-        subprocess.run(["systemctl", "--user", "stop", _SERVICE_NAME])
+        proc = subprocess.run(["systemctl", "--user", "stop", _SERVICE_NAME])
+        if proc.returncode != 0:
+            sys.exit(proc.returncode)
     elif action == "status":
         subprocess.run(["systemctl", "--user", "status", _SERVICE_NAME])
     elif action == "logs":
--- a/v4/tests/test_no_unchecked_subprocess_success.py
+++ b/v4/tests/test_no_unchecked_subprocess_success.py
@@ -40,13 +40,16 @@ _REVIEWED: dict[tuple[str, str], str] = {
         "install's returncode and stays true regardless; a failed patch surfaces as an "
         "unreachable Loki in `kubeintellect status`"
     ),
-    ("cli.py", "systemctl --user start"): (
-        "`service start|stop|status` pass systemctl's own stdout and stderr straight through to "
-        "the terminal — no capture_output, and no ✓ printed afterwards"
+    ("cli.py", "systemctl --user status"): (
+        "`service status` is a query: systemctl may return non-zero for a valid service state "
+        "such as inactive, so its returncode is not treated as a command failure; stdout and "
+        "stderr are passed through directly"
     ),
-    ("cli.py", "systemctl --user stop"): "as `service start`",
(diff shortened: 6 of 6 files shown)


MSKazemi:
@Ryota-Di — first PR to this project, and it is green on all nine required checks. Welcome properly. 🎉

**First, an apology for the delay that was not yours:** your CI had never run. A first-time contributor's fork PR sits at `conclusion: action_required` until a maintainer clicks approve, and `completed` in the API makes it *look* like it ran and failed. That is the trap [@biggdawg320](https://github.com/biggdawg320) documented in [TRIAGE.md](https://github.com/MSKazemi/kubeintellect/blob/main/TRIAGE.md) last week, and I still managed to leave you waiting on it. Approved now, and everything passes.

## I re-ran your red-green rather than taking your word for it

Your claim was that the test fails before the fix and passes after. Verified independently, by putting `main`'s unpatched `cli.py` back underneath your test:

```console
$ pytest tests/test_no_unchecked_subprocess_success.py -k propagate_systemctl_failure
E       Failed: DID NOT RAISE SystemExit
FAILED ...test_service_start_and_stop_propagate_systemctl_failure[start]
FAILED ...test_service_start_and_stop_propagate_systemctl_failure[stop]
2 failed

$ # with your fix restored
17 passed in 3.28s
```

Byte-for-byte what you reported. That matters more than it sounds: an accurate self-report is the thing that makes the *next* PR cheap to review.

## The part that shows you read the code rather than the issue

Anyone can add `if proc.returncode != 0`. Two things here are better than that.

**You propagated the *actual* exit code** — `sys.exit(proc.returncode)`, not `sys.exit(1)`. A caller can still tell a unit-not-found from a permission denial.

**And you noticed the allowlist would rot.** `_REVIEWED` had:

```python
("cli.py", "systemctl --user stop"):   "as `service start`",
("cli.py", "systemctl --user status"): "as `service start`",
```

Deleting the `start` entry — which your fix requires — would have left `status` and `logs` pointing at a reason that no longer existed. A reason nobody can read is the same as no reason, and that allowlist only works because every entry carries one. You rewrote both with real exit-code semantics instead. I checked; nothing dangling is left behind.

**Your call to leave `status` and `logs` alone is correct, and it is the interesting half of this issue.** `systemctl status` returns non-zero for an *inactive* service — a perfectly valid answer to a query. Propagating that would have made `kubeintellect service status` fail whenever it successfully told you the service was stopped. And `journalctl -f` exits non-zero on the user's own Ctrl-C. Treating either as failure would have been a new bug wearing the old bug's clothes. Two of the four sites were deliberately allowlisted and you worked out which two.

You also updated the documented server counts (5562 → 5564) in both `AGENTS.md` and `CONTRIBUTING.md` — that number is gated, so a PR adding tests goes red without it. Most first PRs here trip on exactly that.

## Your row on the wall

You were added to the [contributor table](https://github.com/MSKazemi/kubeintellect#contributors) two days ago marked 🛠️, when you claimed the issue rather than when you delivered. Per the rule announced then, the row gets **rewritten to describe this fix** once it lands — from "stepped forward" to what you actually did.

---

Merge decision is with the maintainer and I expect it to be yes. If you want another, [#154](https://github.com/MSKazemi/kubeintellect/issues/154) is close in spirit — count-drift gates that run per-PR and cannot see cross-PR drift, which is the gate you just satisfied, viewed from the other side. [`good first issue`](https://github.com/MSKazemi/kubeintellect/labels/good%20first%20issue) has the rest. No obligation at all.

And entirely optionally: a ⭐ or a note anywhere about *"a CLI that exited 0 when the service it started had refused to start"* helps a one-maintainer project more than it looks — the repo currently has more forks than stars. It changes nothing about how your work is treated either way. Thank you for a genuinely well-made first contribution. 💙

MSKazemi:
Merged — thank you, this was a real bug and a clean fix.

Your change is on `main` here: https://github.com/MSKazemi/kubeintellect/blob/a11425dc556f3de1a01c287660c65e8d76459cf8/v4/packages/kubeintellect-server/app/cli.py#L782-L789

What it fixes matters more than its size. `service start` exiting 0 after systemd refused is the failure mode this project treats as the worst one: a confident success for work that did not happen. Anything that trusted that exit code — a deploy script, a CI step, an operator reading a terminal — was told the server was up when it was not.

The part I want to call out is the part you *didn't* change. Leaving `service status` and `service logs` discarding their return codes is the correct call, and it is the easy thing to get wrong in this patch: `systemctl status` returns non-zero for a perfectly valid inactive unit, and `journalctl -f` returns non-zero when the operator presses Ctrl-C. Treating either as a failure would have swapped one wrong answer for another. Rewriting the surviving `_REVIEWED` reasons in exit-code terms instead of pointing at the entries you deleted was also right — the dead-entry guard would have caught the alternative.

Verified before merging, on the merged tree: ruff clean, mypy clean across 193 files, server suite 5560 passed, kq suite 749 passed, and the file-mode, syntax, encoding, roster and doc-claims gates all green. I confirmed your test genuinely fails on the unpatched code first (`DID NOT RAISE SystemExit`, both parameters) rather than taking it on faith.

Two things I added on top, so you know exactly what changed under your name:

1. A success-path test. Yours pins the code that gets propagated; nothing held the other side, so `sys.exit(proc.returncode or 1)` — the shape a later cleanup reaches for — would turn every *successful* `service start` into exit 1 with the entire suite still green. I checked that by injecting exactly that line; only the new case fails.
2. `systemctl --user start` and `stop` in `_MUST_STAY_CHECKED`. Deleting the allowlist entries stops the discarded shape returning, but nothing stopped someone writing the entry back, which is the repair a future contributor would reach for first.

I also merged `main` into your branch to resolve the test-count conflict and recollected the number (5578 -> 5584). Your commit is untouched in the history.

Your all-contributors entry is updated from the in-flight placeholder to bug/code/test, and the changelog entry for the next release credits you by name.

If you're looking for another one, the same class of bug is worth hunting elsewhere in `cli.py` — `service logs` still exits 0 when the unit does not exist and `journalctl -f` returns immediately having streamed nothing. That is a separate finding from yours and I'd be glad to see it.

Entirely optional, and the merge and the credit were the point either way: if the project is useful to you, a star or a mention helps other people find it.

PR #210, opened 2026-09-11
feat(playbooks): add evidence-based Ingress TLS expiry triage

biggdawg320:
## What & why

Related to #13; claims only the Ingress TLS/certificate-expiry topic in https://github.com/MSKazemi/kubeintellect/issues/13#issuecomment-5630085867. Leaves the umbrella issue open.

Add `TlsCertificateExpired`, a triage-only playbook for a certificate actually served to an Ingress client. It requires endpoint/termination ownership, SNI/hostname, certificate identity and trusted time before diagnosing expiry. It distinguishes expired intermediates, clock skew, future validity, external termination and a renewed Secret that is not yet served. No Secret dumping, private-key output, verification bypass, automatic renewal or RBAC expansion.

The Go error says "expired or is not yet valid" in both cases: its trigger requires the same-line timestamp and `is after` detail. Other triggers recognize explicit curl/Python/OpenSSL expiry errors. They select a guide when that text is supplied to the matcher; they do not introduce client/log ingestion or an automatic expiry monitor. `detect: null` is intentional.

## Type and scope

- [x] New playbook, documentation and tests in v4; root documentation counts regenerated.
- [x] New tests run red first: 5 failed / 14 passed before adding the playbook.
- [x] Positive expiry cases and negative healthy, future-validity, unknown-CA, hostname-mismatch, generic handshake/503 and cross-line cases.
- [x] Existing human approval and secret-handling boundaries retained. Certificate/Secret changes are guidance for an authorized operator only.

## Validation

- Final 178 focused tests pass on **both Python 3.12 and 3.13**: new TLS tests, existing playbooks/PDB/StatefulSet, all-trigger reachability and doc-claims tests.
- Ruff passes over the exact CI scope; Linux-targeted mypy passes (193 source files).
- File-mode gate passes (963 indexed files, using an LF-normalized copy of the existing gate under Git Bash); syntax and encoding gates pass (569 tracked Python files); roster passes (14 contributors).
- Generated doc counts verified after staging the new tests; MkDocs builds with existing missing-link/anchor warnings. Whitespace checks pass.
- **Full Windows suites are not green.** Python 3.12 server: 80 failed / 5392 passed / 106 skipped / 6 errors. Python 3.13 server: 81 failed / 5391 passed / 106 skipped / 6 errors. CLI: 10 failed / 739 passed on each interpreter. Server failure/error names match the previously recorded Windows baseline except the 3.13 doc-count failure; that count was regenerated after adding the new test file to the index and passes in both final focused runs. CLI failure names match the previous baseline. This is comparison against recorded baseline logs, not a fresh unchanged-base run. No full-suite rerun after the final documentation-count correction.
- Local staged tree exactly matches the published tree `9b88ff7`.

## Notes for reviewers

No live cluster, endpoint probe, LLM call or paid service used. Routing tests verify candidate selection, not a production expiry diagnosis. Source references: https://go.dev/src/crypto/x509/verify.go and https://kubernetes.io/docs/concepts/services-networking/ingress/#tls .

Prepared and validated with substantial OpenAI Codex assistance under the account owner's direction.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -91,7 +91,7 @@ uv run ruff check packages/kubeintellect-server/app/ packages/ki-protocol/ packa
 # 2. Types — the workspace is at ZERO errors; keep it there.
 uv run mypy packages/kubeintellect-server/app packages/ki-protocol packages/kube-q/kube_q
 
-# 3. Server suite (5562 tests)
+# 3. Server suite (5584 tests)
 uv run python -m pytest tests/ -q
 
 # 4. kq CLI suite (749 tests)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -204,7 +204,7 @@ uv run mypy packages/kubeintellect-server/app packages/ki-protocol packages/kube
 
 # 3. Tests — two suites, run separately: the two `tests` packages collide
 #    under a single pytest invocation.
-uv run python -m pytest tests/ -q                              # server (~5562 tests)
+uv run python -m pytest tests/ -q                              # server (~5584 tests)
 cd packages/kube-q && uv run python -m pytest tests/ -q        # kq CLI (~749 tests)
 ```
 
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -29,7 +29,7 @@ The `v4/` platform is the current line and the only one you should deploy.
 - Conversational diagnosis correlating **kubectl + Prometheus + Loki**.
 - **Human-in-the-loop approval gate** with RBAC (superadmin / admin / operator / readonly) on
   every mutating action, enforced server-side at the mutating chokepoint.
-- **26 declarative failure playbooks** (`detect → investigate → remediate`), 20 of which compile
+- **27 declarative failure playbooks** (`detect → investigate → remediate`), 20 of which compile
   to zero-token detectors.
 - **Zero-token detection** — a detector engine over a kubectl `--watch` sensorium fires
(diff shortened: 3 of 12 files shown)


MSKazemi:
Thanks, @biggdawg320 — this is in: [275e41d9](https://github.com/MSKazemi/kubeintellect/commit/275e41d9c23d85e375019ab8a366677d192e871c), shipping in the next release.

Your playbook is the one that ships. I tested it against the real 27-playbook registry alongside @RahulSinha9's independent #211 (filed a day after yours) on five snapshots: yours fired only on genuine `x509: certificate ... is after` expiry evidence and stayed silent on CrashLoopBackOff, ImagePullBackOff, a readiness-probe 500, and a FailedScheduling case — #211's broader event-reason match fired on all four of those too, which would have injected TLS guidance into unrelated incidents. That precision is exactly what made yours the base.

I did fold in three things from #211 that yours didn't cover — the concrete SNI inspection command, reading validity/SAN out of `tls.crt` without touching `tls.key`, and the case where an unexpired cert still fails on SAN mismatch. Full detail and the measured comparison table are in #226.

Your commit (`0342cab`) is carried unmodified in the merged history, so you're in GitHub's contributor graph for this the normal way — and credited in the CHANGELOG and the playbook's file header.

Really appreciate the evidence-based approach — scoping the trigger to the exact Go TLS error message rather than a broad event-reason match is the kind of precision this project needs more playbooks to have.

Closing this in favor of #226, where it's merged.

PR #211, opened 2026-09-11
feat(playbooks): add TLS certificate expiry failure playbook

RahulSinha9:
## Summary

Implements #13 by adding a Kubernetes TLS/certificate-expiry failure playbook and focused tests.

### Changes
- Add `TLSCertificateExpiry` playbook with triggers for common x509/TLS certificate failures.
- Add investigation steps covering pod events/logs, TLS Secrets, cert-manager resources, external endpoints, SAN/hostname validation, and trust chains.
- Keep remediation human-approved and explicitly avoid exposing private keys.
- Add focused matching/schema/negative tests.
- Update the playbook inventory guard from 26 → 27 playbooks and 46 → 48 trigger regexes.

### Validation
- New tests cover positive certificate-expiry matching, schema completeness, and a healthy-event negative case.
- Inventory guard updated as required for a new playbook.

Signed-off-by: Rahul Kumar <krahul088176@gmail.com>

### DIFF
--- a/v4/CONTRIBUTING_TMP.txt
+++ b/v4/CONTRIBUTING_TMP.txt
@@ -1,0 +1,1 @@
+placeholder
--- a/v4/CONTRIBUTING_TMP.txt
+++ b/v4/CONTRIBUTING_TMP.txt
@@ -1,1 +1,0 @@
-placeholder
--- a/v4/packages/kubeintellect-server/app/agent/playbooks/tls_certificate_expiry.yaml
+++ b/v4/packages/kubeintellect-server/app/agent/playbooks/tls_certificate_expiry.yaml
@@ -1,0 +1,30 @@
+name: TLSCertificateExpiry
+triggers:
+  - event_reason_regex: "Failed|Unhealthy|BackOff"
+  - event_message_regex: "x509: certificate.*expired|certificate has expired|certificate verify failed|tls: failed to verify certificate|certificate signed by unknown authority|SSL: CERTIFICATE_VERIFY_FAILED"
+investigation_steps:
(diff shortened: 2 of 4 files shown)


MSKazemi:
Thanks, @RahulSinha9 — the concrete SNI/SAN inspection you added (`openssl s_client … -showcerts`, reading validity and SAN coverage out of `tls.crt` without touching `tls.key`, and covering the case where an unexpired cert still fails because its SAN doesn't cover the requested hostname) is now shipped in [275e41d9](https://github.com/MSKazemi/kubeintellect/commit/275e41d9c23d85e375019ab8a366677d192e871c) — the parts of this playbook the project didn't have.

The base playbook that ships is @biggdawg320's #210, filed the day before yours — I tested both against the real 27-playbook registry on five snapshots, and #210's trigger (scoped to the exact same-line "is after" detail in Go's combined expired-or-not-yet-valid TLS error) fired only on genuine certificate expiry. Your broader `Failed|Unhealthy|BackOff` event-reason match fired on four unrelated cases too (CrashLoopBackOff, ImagePullBackOff, a readiness-probe 500, FailedScheduling) because it searches the whole events blob rather than a reason field — that would have injected TLS guidance into incidents that have nothing to do with certificates.

One thing I deliberately left out: the `kubectl get certificate,certificaterequest` step. The shipped Helm chart's ClusterRole doesn't grant anything under `cert-manager.io`, so it would fail on its own first command — widening that role is a security decision, not something to bundle into a playbook fix.

You're credited in `.all-contributorsrc`, the README contributor table, and the CHANGELOG entry for this. Full comparison and reasoning: #226.

Appreciate you building the SAN/SNI-inspection path — that's real depth this playbook was missing.

Closing this in favor of #226, where your work is merged.

PR #213, opened 2026-09-11
fix(encoding): subprocess.run(..., text=True) decoded with the locale encoding too

### LINKED ISSUE #168: subprocess.run(..., text=True) decodes with the locale encoding too — 23 sites the new encoding gate cannot see

1cbyc:
Closes #168.

`subprocess.run(..., text=True)` decodes the child's pipes with the platform default.
Structured or identifying output now decodes strictly (`encoding="utf-8"`); free-form human-read text uses `errors="replace"`.

### DIFF
--- a/scripts/check-contributor-roster.py
+++ b/scripts/check-contributor-roster.py
@@ -63,7 +63,7 @@ def repo_root() -> str:
     return subprocess.run(  # noqa: S603
         ["git", "rev-parse", "--show-toplevel"],  # noqa: S607
         capture_output=True,
-        text=True,
+        text=True, encoding="utf-8",
         check=True,
     ).stdout.strip()
 
--- a/scripts/check-required-checks.py
+++ b/scripts/check-required-checks.py
@@ -60,7 +60,7 @@ def live_protection() -> dict | None:
     try:
         out = subprocess.run(
             ["gh", "api", f"repos/{REPO}/branches/main/protection"],
-            capture_output=True, text=True, timeout=30, check=False,
+            capture_output=True, text=True, encoding="utf-8", timeout=30, check=False,
         )
     except (FileNotFoundError, subprocess.SubprocessError):
         return None
--- a/scripts/check-syntax-warnings.py
+++ b/scripts/check-syntax-warnings.py
@@ -61,7 +61,7 @@ def repo_root() -> str:
         ["git", "rev-parse", "--show-toplevel"],
         capture_output=True,
         check=True,
-        text=True,
+        text=True, encoding="utf-8",
     ).stdout.strip()
 
 
@@ -81,7 +81,7 @@ def tracked_python_files(root: str | None = None) -> list[str]:
         ["git", "ls-files", "-z", "--", "*.py"],
         capture_output=True,
         check=True,
-        text=True,
+        text=True, encoding="utf-8",
         cwd=root,
     ).stdout
     paths = [p for p in out.split("\0") if p]
--- a/scripts/check-text-encoding.py
+++ b/scripts/check-text-encoding.py
@@ -36,8 +36,14 @@
 # It DOES flag a source file that is not valid UTF-8, rather than skipping it: a
 # file the gate cannot read is a file the gate is not guarding.
 #
-# NOT COVERED, and tracked separately: `subprocess.run(..., text=True)` decodes
-# with the locale encoding too. Same bug class, different call shape.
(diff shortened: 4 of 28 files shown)


1cbyc:
okay, i have concerns. what's wrong?

MSKazemi:
@1cbyc — nothing is wrong on your end, sorry for the scare. That message right above yours is the automated first-PR notice, and the specific thing it's warning about is exactly what you're seeing: GitHub blocks CI on a fork PR from a first-time contributor until a maintainer approves the run. I've now approved it twice (once for your original commit, once just now for the one below), so it should start running shortly.

I reviewed the fix and ran the full gate suite locally against your commit (`391990d`): `ruff check` clean, `mypy` 0 errors across 193 files, server suite 5538 passed / 23 skipped (the one unrelated failure is a local-environment quirk on my machine, not your code), kube-q suite 749 passed, `check_doc_claims.py` passes. The `errors="strict"` vs `errors="replace"` split you and I settled on #168 is implemented correctly everywhere you applied it.

I found one gap and fixed it myself rather than sending it back to you: three call sites read free-form content the same way `run_kubectl`'s `logs`/`describe`/`get` verbs do — `context_fetcher._kubectl_snapshot` (the cluster-health snapshot the LLM reads), `gitops._default_runner`, and `helm_tool.run_helm` — but were left `errors="strict"` only. Each is wrapped in a broad exception handler, so it wasn't crashing, but a single non-ASCII byte there would have collapsed the whole snapshot or helm call to "unavailable" instead of degrading one glyph, which is the wrong side of the line you drew everywhere else. I pushed a follow-up commit (`f2616ab`) on top of yours that adds `errors="replace"` to those three, with your and insisong's `Co-authored-by:` intact — full diff and reasoning in the commit message.

Everything else is unchanged and stays exactly as you wrote it. Thank you for taking #168 — the content-vs-identifier framing you proposed when claiming it was the right call, and it's the version that shipped.

